PR #23, opened 2019-06-06
Update inbox_carousel_layout.xml

faisalkhan1690:
For the id that you are using before defining it, you need to add + sign, because your component doesn't know that id, isn't defined yet.


darshanclevertap:
I have made changes as per this PR myself. We couldn't merge this directly because of our release plans. Thank you so much and keep contributing 🙂

PR #123, opened 2020-11-27
Feature/ph identity/sdk 395

kumaratulismu:


clevertap-core/src/main/java/com/clevertap/android/sdk/ManifestInfo.java
```diff
@@ -90,12 +93,28 @@ private ManifestInfo(Context context) {
...
     public String getFCMSenderId() {
         return fcmSenderId;
     }
 
+    public String[] getProfileKeys() {
+        return profileKeys;
+    }
+
+    private String[] parseProfileKeys(final Bundle metaData) {
+        String profileKeyString = _getManifestStringValueForKey(metaData, Constants.CLEVERTAP_IDENTIFIER);
```
darshanclevertap, line 113: `_getManifestStringValueForKey` returns NULL when the code flow goes into `catch`. Because of that we need to add a null check in the next line

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapInstanceConfig.java
```diff
@@ -1,5 +1,6 @@
 package com.clevertap.android.sdk;
```
darshanclevertap, line 2: File is not formatted.

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapInstanceConfig.java
```diff
@@ -93,9 +73,20 @@ public CleverTapInstanceConfig createFromParcel(Parcel in) {
 
     private boolean useGoogleAdId;
 
+    public void setProfileKeys(@IdentityType String... profileKeys) {
+        if (!isDefaultInstance) {
```
darshanclevertap, line 77: We don't need to check if it's a default instance because this method will be called before the instance creation. At that time, `isDefaultInstance` will be null

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
```diff
@@ -4195,20 +4291,22 @@ private void _onUserLogin(final Map<String, Object> profile, final String clever
             }
 
             boolean haveIdentifier = false;
-
+            LoginInfoProvider handler = new LoginInfoProvider(this);
```
darshanclevertap, line 4294: rename `handler` to `loginInfoProvider`

clevertap-core/src/main/java/com/clevertap/android/sdk/login/LegacyIdentityRepo.java
```diff
@@ -0,0 +1,35 @@
...
+    private final BaseCTApiListener mCTApiListener;
+
+    private final IdentitySet identities;
+
+    public LegacyIdentityRepo(final BaseCTApiListener ctApiListener) {
+        this.mCTApiListener = ctApiListener;
+        this.identities = IdentitySet.getDefault();
+    }
+
+    @Override
```
darshanclevertap, line 23: Add javadoc

clevertap-core/src/main/java/com/clevertap/android/sdk/ManifestInfo.java
kumaratulismu: check is added to prevent null check

darshanclevertap (approved):
Looks good to me 👍

PR #173, opened 2021-05-03
Task/GitHub issues/sdk 778

william-ct:
This PR contains fixes added for console and GitHub issues.Following are issues which are fixed:-
CTBackgroundJobService crash on Android 4.4 SDK-779
NPE on InstallReferrer code silent crash SDK-780
Notification Enabled Crash on SDK Console SDK-781

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/DeviceInfo.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/DeviceInfo.java
@@ -12,6 +12,7 @@ import android.content.res.Configuration;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.os.Build;
+import android.os.RemoteException;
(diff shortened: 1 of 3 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/DeviceInfo.java
```diff
@@ -186,7 +187,13 @@ private String getNetworkType() {
         }
 
         private boolean getNotificationEnabledForUser() {
-            return NotificationManagerCompat.from(context).areNotificationsEnabled();
+            boolean isNotificationEnabled;
```
darshanclevertap, line 190: modify boolean value here only if we get a genuine false. Catch exception and print logs

darshanclevertap (approved):
Looks good @william-ct 👍🏾

PR #184, opened 2021-06-18
fix record screen NPE SDK-803

piyush-kukadiya:
- fix record screen NPE
- add record screen button in sample app


darshanclevertap (approved):
@piyush-kukadiya LGTM 👍🏾

PR #186, opened 2021-06-18
SDK-886): Added code to exclude activity from lifecycle callbacks

darshanclevertap:
- Adds code to give the option of excluding activities from ActivityLifecycleCallbacks


piyush-kukadiya (approved):
LGTM

PR #187, opened 2021-06-20
SDK-886: Add option to exclude from lifecycle callbacks

darshanclevertap:
Allows setting of the app to foreground and setting of the current activity to ensure InApps and Inbox Activities are not affected.


piyush-kukadiya (commented):
`setInitialAppEnteredForegroundTime` logic is missing, Is this intentional?

darshanclevertap:
Closing this as it is not required anymore

PR #192, opened 2021-07-08
async_device id/SDK-934

piyush-kukadiya:
- Added overloaded method of `getCleverTapID()` which returns ID on callback on main thread
- moved `onInitDeviceInfo()` to IO thread
- removed duplicate `notifyUserProfileInitialized()` from `instanceWithConfig()`
- Reinitialised product config & Feature Flag controllers with device id if it's null during first initialisation


darshanclevertap (approved):
LGTM 👍🏾

PR #194, opened 2021-07-15
task(async_deviceID): remove strict mode logging SDK-934

william-ct:
Removed strict mode setting code from sample app


darshanclevertap (approved):
LGTM

PR #208, opened 2021-11-24
Fix/strict mode issue(sdk 1163)

william-ct:
*Fixes a Strict Mode Read violation(getDeviceID()) that was causing ANR


darshanclevertap (approved):
Looks good 👍🏾

PR #253, opened 2022-03-31
task(SDK-1431): expose callback to notify about the DC domain availability

shivamsharma2710:


clevertap-core/src/main/java/com/clevertap/android/sdk/network/NetworkManager.java
```diff
@@ -237,14 +238,18 @@ public void incrementResponseFailureCount() {
     @Override
     public void initHandshake(final EventGroup eventGroup, final Runnable handshakeSuccessCallback) {
         responseFailureCount = 0;
-        setDomain(context, null);
+        //setDomain(context, null);
```
darshanclevertap, line 241: @shivamsharma2710 Why is this commented? Needed or can be removed?

shivamsharma2710: @darshanclevertap This was intentionally commented. 



### Explanation: 
Commented line was calling the `onDCDomainUnavailable()` even for the initial handshake which is not the valid use of the DC callback and was creating the issue in handling.   

To get the domain, SDK calls the initHandshake() in following two scenarios,

1. To perform the **initial handshake**, when the regionCode is not defined and the domain is null
    (This is the case of primary handshake)
    
2. When the max retryLimit of 5 attempts exceeds while sending the event queue

To fulfil the both Core and DC requirements, commented line was moved into the`needsHandshakeForDomain()` so SDK started reporting the domain unavailability in the valid case.

darshanclevertap: @shivamsharma2710 If this needs to be commented then just remove it.

shivamsharma2710: @darshanclevertap done.

darshanclevertap (approved):

PR #291, opened 2022-07-12
[SDK-1704] [SDK-1705] jacoco modifications

root-ansh-ct:
- modify test coverage to run only when passed via build arguments


piyush-kukadiya (approved):
LGTM

PR #325, opened 2022-09-12
fix(push-templates): add subtitle to notification header above android 12 SDK-2160

piyush-kukadiya:
- fixes a bug on android 12 where push template notification header was not displaying subtitle text.


darshanclevertap (approved):

MrPiNkY2022 (changes requested):
Opinions please???

PR #362, opened 2022-11-22
[SDK-2293] multiple  NOC fixes

root-ansh-ct:
- add Proguard rules to keep CREATOR instance for Parcelable classes  (fixes ANR 1 from SUC-119607) 
- make call to `installReferral` async   (fixes crash 3 from SUC-119607) 
- use concurrent HashMap instead of HashMap for storing `CleverTapApi` instances (fixes crash 1 and 2 from SUC-119607) 
- update unit tests


piyush-kukadiya (approved):
LGTM

PR #370, opened 2022-12-05
fix ANR SDK-2328

piyush-kukadiya:
ANR was causing due to static initialization of SimpleDateFormate on main thread. It has been removed since unused in our code.

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/Constants.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/Constants.java
@@ -5,10 +5,8 @@ import androidx.annotation.RestrictTo;
 import androidx.annotation.StringDef;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
-import java.text.SimpleDateFormat;
 import java.util.Arrays;
 import java.util.HashSet;
-import java.util.Locale;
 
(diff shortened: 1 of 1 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/Constants.java
```diff
@@ -49,7 +47,6 @@ public interface Constants {
     int SESSION_LENGTH_MINS = 20;
     String DEVICE_ID_TAG = "deviceId";
     String FALLBACK_ID_TAG = "fallbackId";
-    SimpleDateFormat FB_DOB_DATE_FORMAT = new SimpleDateFormat("MM/dd/yyyy", Locale.US);
```
darshanclevertap, line 52: @piyush-kukadiya this variable was unused?

piyush-kukadiya: @darshanclevertap YES. It's Facebook profile DOB related code which we removed in the past when removing google and facebook login support

darshanclevertap: Ok thanks!

darshanclevertap (approved):
👍🏾

PR #410, opened 2023-03-31
[SDK-2851]: App Inbox parity for itemIndex parameter of the on onInboxItemClicked callback

shivamsharma2710:
Changed the value that we pass in the `itemIndex` parameter of the `onInboxItemClicked` callback.

Earlier it was the position of the item clicked from the list container. Now, it is the index of the content-page. 
Default value for all the templates is 0 and it varies from 0 to number of pages for carousel templates.


piyush-kukadiya (approved):
LGTM

darshanclevertap (approved):

PR #411, opened 2023-04-04
task(SDK-2575): Fix retry mechanism bug for non-EU1 regions

darshanclevertap:
- Makes the `minDelayFrequency` variable global to the class so that the retry mechanism logic works as intended
- Cleans up code


piyush-kukadiya (approved):
LGTM

PR #416, opened 2023-04-07
[SDK-2829] App Inbox changes which Android v4.6.9 supports

shivamsharma2710:
* **Signature change of `onInboxItemClicked` callback**:
  It is changed from  `onInboxItemClicked(CTInboxMessage message)` to `onInboxItemClicked(CTInboxMessage message, int contentPageIndex, int buttonIndex)`. The `contentPageIndex` corresponds to the page index of the content, which ranges from 0 to the total number of pages for carousel templates. For non-carousel templates, the value is always 0, as they only have one page of content. The `buttonIndex` corresponds to the App Inbox button clicked (0, 1, or 2). A value of -1 indicates the App Inbox item is clicked.

* **Behavioral change of `onInboxItemClicked` callback**:
  - Previously, the callback was raised when the App Inbox Item is clicked.
  - Now, it is also raised when the App Inbox button and Item is clicked.

#### Added
* Adds the new public API `dismissAppInbox()` via `CleverTapAPI` class to dismiss the App Inbox.

### DIFF
--- a/docs/EXAMPLES.md
+++ b/docs/EXAMPLES.md
@@ -233,7 +233,7 @@ Lets's understand the types of buttons first that App Inbox supports:
 
 
 The Android SDK v4.6.1 and above supports `onInboxItemClicked` callback on the click of an App Inbox item, such as text or media.
-From the Android SDK v4.7.6, the `onInboxItemClicked` callback supports the button click besides the item click.
+From the Android SDK v4.6.8 onwards and below v4.7.0, the `onInboxItemClicked` callback supports the button click besides the item click.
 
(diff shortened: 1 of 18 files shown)


darshanclevertap:
@shivamsharma2710 please check why lint checks are failing

docs/EXAMPLES.md
```diff
@@ -283,7 +283,7 @@ cleverTapDefaultInstance.initializeInbox();
 //Get Inbox Message Count
 cleverTapDefaultInstance.getInboxMessageCount();
 
-//Get Inbox Unread Count
+//Get Inbox Unre ad Count
```
darshanclevertap, line 286: Unread is one word @shivamsharma2710

shivamsharma2710: this is done.

PR #417, opened 2023-04-10
[SDK-2901] Fix PE flag for development mode.

hborisoff:
When using the `BuildConfig.DEBUG` flag the development mode is not working, because when a release version of the SDK is built this flag gets replaced by `true` or `false` value. Using the flag directly from the context fixes it.


darshanclevertap (approved):

piyush-kukadiya (approved):
LGTM

PR #430, opened 2023-06-09
[SDK-2955]: Version Catalog 

william-ct:
Replaces refreshVersions plugin with Version catalog of Gradle.


build.gradle
```diff
@@ -10,16 +10,14 @@ buildscript {
...
     }
     dependencies {
-        classpath Libs.com_android_tools_build_gradle
-        classpath Libs.google_services// Google Services plugin
-        //classpath "com.github.dcendents:android-maven-gradle-plugin:$mavenPluginVersion"
-        classpath Libs.agcp// Huawei Push Plugin
-        classpath Libs.kotlin_gradle_plugin
-        classpath Libs.sonarqube_gradle_plugin
-        classpath 'com.google.firebase:firebase-crashlytics-gradle:2.8.1'
+        classpath("com.google.firebase:firebase-crashlytics-gradle:2.8.1")
```
darshanclevertap, line 13: @william-ct why do we have crashlytics in our dependencies?
Also why isn't `detekt` library part of the version catalog?

test_shared/build.gradle
```diff
@@ -26,33 +26,33 @@ android {
...
-    api Libs.kotlin_stdlib_jdk7
-    api Libs.jsonassert
-    api Libs.gson
+    api (libs.test.junit.platform.runner)
+    api (libs.test.junit.jupiter.api)
+    api (libs.test.junit.jupiter.engine)
+    api (libs.kotlin.stdlib.jdk7)
+    api (libs.test.jsonassert)
+    api (libs.gson)
     def mockito_version = '3.5.11'
```
darshanclevertap, line 35: @william-ct check if we can move this version catalog as well?

versions.properties
```diff
@@ -1,385 +1,26 @@
...
-##                  # available=1.1.4-alpha06
-##                  # available=1.1.4-alpha07
-##                  # available=1.1.4-beta01
-version.com.google.code.gson..gson=2.8.6
-##                     # available=2.8.7
-##                     # available=2.8.8
-##                     # available=2.8.9
-##                     # available=2.9.0
-##                     # available=2.9.1
+plugin.android=7.4.2
```
darshanclevertap, line 1: @william-ct why do we still need this file?

test_shared/build.gradle
Anush-Shand: Done

build.gradle
Anush-Shand: The sample app uses the crashanalytics plugin. The dependency can be moved to the required gradle file as well
I think William decided to keep libraries only used once away from the versions.toml file. Can be changed

build.gradle
```diff
@@ -36,8 +35,6 @@ allprojects {
     }
 }
 
-apply plugin: 'jacoco'
```
CTLalit, line 39: we removed this and did not ad anywhere, intentional?

Anush-Shand: It was currently redundant. We already apply it in commons.gradle file

CTLalit (approved):

Issue #428, opened 2023-05-31
Race-condition when detecting if an in-app message should show


henryglendening-hh:
**Describe the bug**
There seems to be a race-condition that may be experienced in regards to showing/not showing in-app messages in activities when the `CLEVERTAP_INAPP_EXCLUDE` AndroidManifest key is used. 

In the function `checkExistingInAppNotifications(Activity activity)` of `InAppController.java`, before the SDK attempts to display an in-app message, it first performs a check to ensure that the given activity is not a part of the `CLEVERTAP_INAPP_EXCLUDE` exclusion list. Although this function (`checkExistingInAppNotifications(Activity activity)`) accepts an argument for which activity is being used, the function that it calls to check for exclusion (`canShowInAppOnActivity`) does not accept this argument and instead checks against the _current_ activity. This function works by looping over each excluded activity name, and within each iteration of the loop, the code will do another fetch for the _current_ activity.

These are the two places where the race condition may happen (using the _current_ activity when checking the exclusion list, and using an updated _current_ activity name within each loop iteration). As the app is continuing to transition between activities while the app is started or resumed, the _current_ activity may be quickly changing. Since the `canShowInAppOnActivity` is not referencing the same activity name that `checkExistingInAppNotifications(Activity activity)` pertains too, there's no guarantee (and actually a likelihood) that `canShowInAppOnActivity` will return a value that is inaccurate, leading to an in-app message being shown on an excluded activity.

**To Reproduce**
1. Add the app's launch activity's name to the `CLEVERTAP_INAPP_EXCLUDE`'s value in the AndroidManifest.xml
2. Setup the launch activity to transition to a different activity after 5 seconds
3. Run the app and background it
4. Send a test in-app message from CleverTap to the device
5. Tap on the push notification
6. View the in-app message appear in 1 of 2 places:
- The launch activity
- The other activity

Repeat several times and observe varying results in step 6

**Expected behavior**
The in-app message is not displayed on an excluded activity and is withheld until a different activity is presented

**Environment (please complete the following information):**
 - Device: Samsung Galaxy S20
 - OS: Android 13
 - CleverTap SDK Version: 4.6.1
 - Android Studio Version: 2022.2.1 Patch 2

**Additional context**
Although my app is not running the latest version of the CT SDK, it does appear that the referenced file & functions are using the same code as the version that I'm running (4.6.1), so this bug appears to still affect the latest versions as well

henryglendening-hh:
[...]

PR #433, opened 2023-06-13
fix(SDK-3041): Adds a check to prevent in-app on excluded activity

Anush-Shand:
https://wizrocket.atlassian.net/browse/SDK-3041
https://github.com/CleverTap/clevertap-android-sdk/issues/428
Adds an additional check just before displaying the in-app to prevent the in-app from showing on an excluded activity


piyush-kukadiya (approved):
LGTM

PR #434, opened 2023-06-13
[SDK-2728] Leanplum Interface

hborisoff:
Main interface is `LeanplumCT`, which implements most of the methods of `Leanplum` class from Leanplum SDK.
It can be initialized using the following code:
```
ActivityLifecycleCallback.register(this);
LeanplumCT.initWithContext(this);
```
or by using `CleverTapAPI` instance:
```
ActivityLifecycleCallback.register(this);
LeanplumCT.initWithInstance(instance);
```

Some of the transformation code from CTWrapper in Leanplum repo has been reused.
Most use cases are handled by unit tests.
There are some notes in the [doc](https://docs.google.com/document/d/1639NoXp98ZXCyT1uuIhoVPXB_4dm2U95QylbswRdjNQ/edit).

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/leanplum/CleverTapProvider.kt
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/leanplum/CleverTapProvider.kt
@@ -3,16 +3,14 @@ package com.clevertap.android.sdk.leanplum
 import android.content.Context
 import com.clevertap.android.sdk.CleverTapAPI
 import com.clevertap.android.sdk.Logger
-import java.lang.ref.WeakReference
 
 internal class CleverTapProvider {
 
-  private var contextRef: WeakReference<Context>? = null
   private var customInstance: CleverTapAPI? = null
+  private var defaultInstance: CleverTapAPI? = null
 
   constructor(context: Context) {
-    this.contextRef = WeakReference(context)
-    CleverTapAPI.getDefaultInstance(context) // create default instance
+    this.defaultInstance = CleverTapAPI.getDefaultInstance(context)
   }
 
   constructor(customInstance: CleverTapAPI) {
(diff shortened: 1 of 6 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/leanplum/CleverTapProvider.kt
```diff
@@ -0,0 +1,35 @@
...
+    this.customInstance = customInstance
+  }
+
+  fun getCleverTap(): CleverTapAPI? {
+    if (customInstance != null) {
+      return customInstance
+    } else {
+      val context = contextRef?.get()
+      if (context != null) {
+        return CleverTapAPI.getDefaultInstance(context)
```
piyush-kukadiya, line 28: @hborisoff I would suggest to cache default instance in var, like customInstance. to avoid overhead due to extra call.

hborisoff: @piyush-kukadiya The default instance is not exposed by the Leanplum interface, it is used to forward events to CT, so I don't think it is an overhead, neither for the customers, neither for us. Also if for some reason the instance is changed in the CT SDK, not sure if possible, it will still continue to function properly.

piyush-kukadiya: Each time `ctProvider.getCleverTap()?.` gets called, each time we do look up inside  `CleverTapAPI.getDefaultInstance(context)`. Also if multiple CTWrapper API gets called on different threads then it might lead to duplicate instance creation as `CleverTapAPI.getDefaultInstance(context)` is not thread safe.

piyush-kukadiya: Rest looks good.

hborisoff: @piyush-kukadiya
I fixed it as you suggested. Both references (`customInstance`, `defaultInstance`) in provider class are left for easier visibility when debugging, but only one will be used at a time.

Also the methods calling the constructor of `CleverTapProvider` is almost guaranteed to run on UI thread, we will suggest clients to initialize the LeanplumCT logic in UI thread in the `Application.onCreate` method.

piyush-kukadiya (approved):
LGTM

PR #436, opened 2023-06-27
[SDK-3064] Track events raised from webView interface

shivamsharma2710:
- Passes a flag in the appFields to capture the external initialization/usage of webViewInterface done from application.


piyush-kukadiya (approved):
LGTM

PR #465, opened 2023-08-30
task(SDK-3215) - Updates Logger implementation to static for PT module

Anush-Shand:
Updates Logger implementation to static for PT module
The deprecation is such that the PT loglevel is still respected if set. If it is not set, logs are displayed according to the ClevertapAPI.debugLevel
In the next version, plan is to remove PT logging completely


clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/TemplateRenderer.kt
```diff
@@ -252,7 +251,7 @@ class TemplateRenderer : INotificationRenderer, AudibleNotification {
                         try {
                             ptJsonObj = JSONObject(ptJsonStr)
                         } catch (e: Exception) {
-                            Logger.v("Unable to convert JSON to String")
+                            PTLog.verbose("Unable to convert JSON to String")
```
darshanclevertap, line 254: @Anush-Shand we need to use the Logger of the core SDK here? please confirm with @piyush-kukadiya

PR #468, opened 2023-09-06
task(SDK-3220) - Gets locale info and pass it to LC for LP parity

Anush-Shand:
https://wizrocket.atlassian.net/browse/SDK-3220
Adds an API to setCustomLocale for LP parity
If locale it is not custom set by client then it is fetched from the device and passed to the LC

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
@@ -136,6 +136,8 @@ public class CleverTapAPI implements CTInboxActivity.InboxActivityListener {
 
     private static int debugLevel = LogLevel.INFO.intValue();
 
+    private static String customLocale = null;
+
     static CleverTapInstanceConfig defaultConfig;
 
     private static HashMap<String, CleverTapAPI> instances;
@@ -726,6 +728,27 @@ public class CleverTapAPI implements CTInboxActivity.InboxActivityListener {
     @SuppressWarnings({"unused"})
     public static void setDebugLevel(LogLevel level) {
         debugLevel = level.intValue();
+    }
+
+    /**
+     * Use this method to set a custom locale for the CleverTapAPI
+     *
+     * @param locale - The custom locale to be set
+     */
+    @SuppressWarnings({"unused"})
+    public static void setLocale(String locale) {
+        if(TextUtils.isEmpty(locale))
+            Logger.i("Empty Locale provided for setCustomLocale");
+        customLocale = locale;
+    }
+
+    /**
+     * Returns the custom locale set for CleverTapAPI
+     *
+     * @return The customLocale string value
+     */
+    public static String getLocale() {
+        return customLocale;
     }
 
     /**
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
@@ -3391,8 +3391,10 @@ public class CleverTapAPI implements CTInboxActivity.InboxActivityListener {
      */
     @SuppressWarnings({"unused"})
     public void setLocale(String locale) {
-        if(TextUtils.isEmpty(locale))
-            Logger.i("Empty Locale provided for setCustomLocale");
+        if(TextUtils.isEmpty(locale)) {
+            Logger.i("Empty Locale provided for setLocale, not setting it");
+            return;
+        }
         coreState.getDeviceInfo().setCustomLocale(locale);
     }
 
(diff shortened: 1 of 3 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
```diff
@@ -728,6 +730,27 @@ public static void setDebugLevel(LogLevel level) {
...
+    /**
+     * Use this method to set a custom locale for the CleverTapAPI
+     *
+     * @param locale - The custom locale to be set
+     */
+    @SuppressWarnings({"unused"})
+    public static void setLocale(String locale) {
+        if(TextUtils.isEmpty(locale))
+            Logger.i("Empty Locale provided for setCustomLocale");
+        customLocale = locale;
```
darshanclevertap, line 742: @Anush-Shand instead of setting `customLocale` as a variable in this class, you can expose a setter inside `DeviceInfo` class
This way you can get the Locale from `DeviceInfo` inside the `CTJsonConverter` class

clevertap-core/src/main/java/com/clevertap/android/sdk/utils/CTJsonConverter.java
```diff
@@ -90,6 +91,13 @@ public static JSONObject from(DeviceInfo deviceInfo, CoreMetaData coreMetaData,
             evtData.put("hgt", deviceInfo.getHeight());
             evtData.put("dpi", deviceInfo.getDPI());
             evtData.put("dt", DeviceInfo.getDeviceType(deviceInfo.getContext()));
+
+            // If locale is set by the client then use that, otherwise fetch it from the device
+            String locale = CleverTapAPI.getLocale();
```
darshanclevertap, line 96: No need to add `CleverTapAPI` as a dependency in this class if you set the locale in DeviceInfo class. Refer to previous comment

darshanclevertap (approved):

darshanclevertap (approved):

clevertap-core/src/main/java/com/clevertap/android/sdk/utils/CTJsonConverter.java
```diff
@@ -90,6 +90,13 @@ public static JSONObject from(DeviceInfo deviceInfo, CoreMetaData coreMetaData,
             evtData.put("hgt", deviceInfo.getHeight());
             evtData.put("dpi", deviceInfo.getDPI());
             evtData.put("dt", DeviceInfo.getDeviceType(deviceInfo.getContext()));
+
+            // If locale is set by the client then use that, otherwise fetch it from the device
+            String locale = deviceInfo.getCustomLocale();
+            if(TextUtils.isEmpty(locale))
+                locale = deviceInfo.getLocale();
```
CTLalit, line 97: if locale is empty we get it from device, but you have created a setter method that allows setting emply locale. Can we make the change in setter itself so empty is not a valid value?

CTLalit (changes requested):
Check comments

clevertap-core/src/main/java/com/clevertap/android/sdk/utils/CTJsonConverter.java
Anush-Shand: We wanted this behaviour similar to leanplum and hence there is no validation on the locale provided in setCustomLocale. Here we expect the user to send an appropriate locale, it can be of any format

clevertap-core/src/main/java/com/clevertap/android/sdk/DeviceInfo.java
```diff
@@ -619,6 +639,23 @@ public void incrementLocalInAppCount() {
...
+        this.customLocale = customLocale;
+    }
+
+    public String getCustomLocale() {
+        return customLocale;
+    }
+
+    public String getLocale() {
+        // If locale is set by the client then use that, otherwise fetch it from the device
+        return TextUtils.isEmpty(getCustomLocale()) ? getDeviceLocale() : getCustomLocale();
```
CTLalit, line 656: This method might return null. Please check

CTLalit (changes requested):
Check comments.

clevertap-core/src/main/java/com/clevertap/android/sdk/DeviceInfo.java
Anush-Shand: If customLocale is null(not set by the user) then we use the deviceLocale which won't be null

CTLalit: TextUtils.isEmpty() -> this one is also checking null internally, resolving, name is misleading.

CTLalit (approved):

PR #476, opened 2023-10-06
task(SDK-3325) - Adds foreground check before initializing XPS

Anush-Shand:
Adds a check to see if app is foreground before initialising xps
https://wizrocket.atlassian.net/browse/SDK-3325


piyush-kukadiya (approved):
LGTM

PR #483, opened 2023-10-16
Build apps

CTLalit:
Instant app + App build optimisations

- Added instant app for same package name so it could be tested
- Jira for instant apps : https://wizrocket.atlassian.net/browse/SDK-3317
- Build files changed to have various flavours we need during release 
- We need following builds:
- local : to build app from local project files
- staging : to build app from staging area of maven (we need to test this before releasing a lib version)
- remote : All deps for CT within the app are pulled in remotely
- Correct build variant can be selected and app can be used


clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
```diff
@@ -937,7 +937,7 @@ private void createAlarmScheduler(Context context) {
             intent.setPackage(context.getPackageName());
 
             int flagsAlarmPendingIntent = PendingIntent.FLAG_UPDATE_CURRENT;
-            if (VERSION.SDK_INT >= VERSION_CODES.M) {
+            if (VERSION.SDK_INT >= VERSION_CODES.S) {
```
darshanclevertap, line 940: @CTLalit if we make this change now then we will need to re-release the core Android SDK as well. Can we exclude this change for now?

sample/build.gradle
```diff
@@ -128,6 +137,18 @@ dependencies {
     implementation "org.jacoco:org.jacoco.core:0.8.4"
     implementation "org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.72"
     implementation "org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:3.1.1"*/
+
+    remoteImplementation("com.clevertap.android:clevertap-android-sdk:5.1.0")
```
darshanclevertap, line 141: @CTLalit should this be update to v5.2.1?

PR #484, opened 2023-10-17
task(SDK-3341) - Fixes star-click Rating PT issue

Anush-Shand:
https://wizrocket.atlassian.net/browse/SDK-3341
Fixes issue on Rating PT where clicking a star led to nothing on API 30 and below
Also addressed the scenario when 2 Rating PT are available at the same time

### DIFF
--- a/clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/content/PendingIntentFactory.kt
+++ b/clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/content/PendingIntentFactory.kt
@@ -355,12 +355,15 @@ internal object PendingIntentFactory {
 
     /**
      * This function returns the number(1 to 5) of the concerned star for rating PT based on the identifier.
-     * The identifiers for the stars are consecutive integers.
-     * Hence the specific star number can be derived by subtracting the first identifier
      */
     @JvmStatic
     private fun getRatingStarNumber(identifier: Int): Int {
-        return identifier - RATING_CLICK1_PENDING_INTENT + 1;
+        return when (identifier) {
+            RATING_CLICK1_PENDING_INTENT -> 1
+            RATING_CLICK2_PENDING_INTENT -> 2
+            RATING_CLICK3_PENDING_INTENT -> 3
+            RATING_CLICK4_PENDING_INTENT -> 4
+            else -> 5
+        }
     }
-
 }
--- a/clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/content/PendingIntentFactory.kt
+++ b/clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/content/PendingIntentFactory.kt
@@ -180,7 +180,7 @@ internal object PendingIntentFactory {
 
             RATING_CLICK1_PENDING_INTENT, RATING_CLICK2_PENDING_INTENT, RATING_CLICK3_PENDING_INTENT,
             RATING_CLICK4_PENDING_INTENT, RATING_CLICK5_PENDING_INTENT -> {
-                val clickedStar = identifier - RATING_CLICK1_PENDING_INTENT + 1
+                val clickedStar = getRatingStarNumber(identifier)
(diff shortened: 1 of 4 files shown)


clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/content/PendingIntentFactory.kt
```diff
@@ -183,71 +178,17 @@ internal object PendingIntentFactory {
                 }
             }
 
-            RATING_CLICK1_PENDING_INTENT -> {
+            RATING_CLICK1_PENDING_INTENT, RATING_CLICK2_PENDING_INTENT, RATING_CLICK3_PENDING_INTENT,
+            RATING_CLICK4_PENDING_INTENT, RATING_CLICK5_PENDING_INTENT -> {
+                val clickedStar = identifier - RATING_CLICK1_PENDING_INTENT + 1
```
CTLalit, line 183: make a function to return **int** based on **identifier** rather than having a magic number function like this.

clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/content/RatingContentView.kt
```diff
@@ -24,6 +26,11 @@ class RatingContentView(context: Context, renderer: TemplateRenderer, extras: Bu
         remoteView.setImageViewResource(R.id.star4, R.drawable.pt_star_outline)
         remoteView.setImageViewResource(R.id.star5, R.drawable.pt_star_outline)
 
+        // Request Codes for all stars are passed as an extra to cancel all pending intents when any of the star is clicked
+        extras.putIntArray(PTConstants.KEY_REQUEST_CODES, IntArray(5) {
+            abs(Random().nextInt())
```
CTLalit, line 31: why do we need the abs() function?

clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/content/RatingContentView.kt
```diff
@@ -24,6 +26,11 @@ class RatingContentView(context: Context, renderer: TemplateRenderer, extras: Bu
         remoteView.setImageViewResource(R.id.star4, R.drawable.pt_star_outline)
         remoteView.setImageViewResource(R.id.star5, R.drawable.pt_star_outline)
 
+        // Request Codes for all stars are passed as an extra to cancel all pending intents when any of the star is clicked
+        extras.putIntArray(PTConstants.KEY_REQUEST_CODES, IntArray(5) {
```
CTLalit, line 30: rather than passing in extras we can pass on the generated int in method signature

CTLalit (changes requested):
Added some comments.

clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/content/RatingContentView.kt
Anush-Shand: Agreed. It is unnecessary. Removing it

clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/content/RatingContentView.kt
Anush-Shand: @CTLalit Do you mean to the getPendingIntent() function? This disturbs the current structure of the code?

CTLalit: We can create a function overload, or have a default value for the param which will not disturb existing code.

Anush-Shand: @CTLalit This can be done but seems like an overkill for now in my opinion. I don't see where else will this new parameter could be needed later

CTLalit (approved):

PR #497, opened 2023-10-30
[SDK-3328] Fix classCastException from while reading PE variable

shivamsharma2710:
**Issue Summary:**
When a PE variable of `Long` type is defined with a null default value, and the dashboard overrides it with a Double value, The SDK throws a ClassCastException (Double cannot be cast to Long).

**Fix:**
Restricted the null usage as default value for PE variables.


Anush-Shand (approved):
LGTM

Anush-Shand (approved):

Anush-Shand:
Unit tests are failing @shivamsharma2710

Anush-Shand (approved):
LGTM

PR #508, opened 2023-11-28
Feat(inapps): chaining -> composition, removal of chaining

CTLalit:
- composed all objects for response into a list
- do not need to chain and pass around multiple refs
- code is simplified and easier to pass on data
- used safe IDE refactor to accomplish


piyush-kukadiya (approved):
LGTM

PR #510, opened 2023-12-05
task(SDK-3422) - Handles NPE for inbox messages

Anush-Shand:
Handles NPE for inbox messages. If an exception was caught earlier when retrieving the messages from DB, a null was returned. Now instead, an empty list is returned


CTLalit (approved):

piyush-kukadiya (approved):
LGTM

PR #514, opened 2023-12-12
Lp/image preload eviction

CTLalit:
Handling eviction of inapps assets.
We need to cleanup the CT SDK downloaded data in form of asssets.
Used a sliding expiry mechanism to check if asset is no longer required and cache is evicted


clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EvaluationManager.kt
```diff
@@ -125,6 +129,11 @@ class EvaluationManager constructor(
                 triggersManager.increment(campaignId)
 
                 val matchesLimits = limitsMatcher.matchWhenLimits(getWhenLimits(inApp), campaignId)
+                val discardData = limitsMatcher.shouldDiscard(getWhenLimits(inApp), campaignId)
+
+                if (discardData) {
+                    clearResource.invoke("") // todo pass correct url
```
piyush-kukadiya, line 135: @CTLalit There is todo. Are we missing url here?

clevertap-core/src/main/java/com/clevertap/android/sdk/StoreProvider.kt
```diff
@@ -56,6 +67,7 @@ class StoreProvider {
 
     fun constructStorePreferenceName(storeType: Int, deviceId: String = "", accountId: String = ""): String =
         when (storeType) {
+            STORE_TYPE_INAPP_ASSETS -> "inapp_assets:$deviceId:$accountId"
```
shivamsharma2710, line 70: @CTLalit we can move `inapp_assets` to the constants

PR #518, opened 2023-12-13
Task/sdk 3514/fallback channel resource name conflict

Anush-Shand:
https://wizrocket.atlassian.net/browse/SDK-3514
Adds ct as prefix to a default channel resource name because of conflict


piyush-kukadiya (approved):
LGTM

PR #522, opened 2023-12-20
task(SDK-3530) - Moves initialisation of ctVariables to an IO thread

Anush-Shand:
StrictMode violation was seen because `ctVariables.init()` led to prefs being accessed on the main thread
https://wizrocket.atlassian.net/browse/SDK-3530


piyush-kukadiya (approved):
LGTM

PR #523, opened 2023-12-20
[WIP] Extract API calls from NetworkManager

vasct:
Extract HTTP calls logic and API definitions from NetworkManager.
Create a simple HttpClient that can be mocked/replaced. That would enable testing of logic that depends on calls and potentially ease the transition to another HttpClient if needed.
Create API class that handles domain changes and API calls

All changes are made with the intent of keeping NetworkManager's interface the same, with only minimal changes to public methods and no changes in actual behavior.


clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/CtApi.kt
```diff
@@ -0,0 +1,112 @@
...
+            }
+        }
+    }
+
+    private fun createRequest(
+        relativePath: String,
+        body: String?,
+        useSpikyDomain: Boolean,
+        includeTs: Boolean
+    ) = Request(getUriForPath(relativePath, useSpikyDomain, includeTs), defaultHeaders, body)
```
CTLalit, line 86: names to call args

clevertap-core/src/main/java/com/clevertap/android/sdk/network/NetworkManager.java
```diff
@@ -694,169 +566,160 @@ public boolean sendQueue(final Context context, final EventGroup eventGroup, fin
             return false;
         }
 
-        HttpsURLConnection conn = null;
-        try {
-            final String endpoint = getEndpoint(false, eventGroup);
+        EndpointId endpointId = EndpointId.fromEventGroup(eventGroup);
```
CTLalit, line 569: please make sure exceptions are not thrown in these blocks as we moved try block

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/CtApi.kt
```diff
@@ -0,0 +1,112 @@
...
+import android.net.Uri
+import com.clevertap.android.sdk.Logger
+import com.clevertap.android.sdk.network.http.CtHttpClient
+import com.clevertap.android.sdk.network.http.Request
+import com.clevertap.android.sdk.network.http.Response
+
+class CtApi(
+    private val httpClient: CtHttpClient,
+    val defaultDomain: String,
+    var domain: String?,
```
CTLalit, line 12: why are these fields defined as var? do they change?

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/CtApiProvider.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.clevertap.android.sdk.network.http.UrlConnectionHttpClient
+
+fun provideDefaultTestCtApi(
+    context: Context,
+    config: CleverTapInstanceConfig,
+    deviceInfo: DeviceInfo
+): CtApi {
+    val httpClient = UrlConnectionHttpClient(config.isSslPinningEnabled, config.logger, config.accountId)
+
+    return CtApi(
```
CTLalit, line 17: names to call args for readability.

CTLalit (commented):
Added some comments

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/CtApi.kt
vasct: `domain` and `spikyDomain` are changed from the corresponding methods of NetworkManager - `setDomain` and `setSpikyDomain`. `region`, `proxyDomain` and `spikyProxyDomain` are currently only used in tests, but from the perspective of the CtApi class I think we should still treat them as possibly changeable in order to be consistent with `domain` and `spikyDomain`.

clevertap-core/src/main/java/com/clevertap/android/sdk/network/NetworkManager.java
vasct: Only getQueueHeader contains lines that can throw and its whole body is currently wrapped with catch(Throwable t). On a different note I don't know why we catch Throwable instead of Exception.

CTLalit: That is legacy code, lets handle the catch block correctly as required.

clevertap-core/src/main/java/com/clevertap/android/sdk/network/http/UrlConnectionHttpClient.kt
```diff
@@ -0,0 +1,101 @@
...
+            val certificateFactory = CertificateFactory.getInstance("X.509")
+            val trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm())
+            val keyStore = KeyStore.getInstance(KeyStore.getDefaultType())
+            keyStore.load(null, null) //Use null InputStream & password to create empty key store
+            val inputStream: InputStream =
+                BufferedInputStream(javaClass.classLoader?.getResourceAsStream("com/clevertap/android/sdk/certificates/AmazonRootCA1.cer"))
+            val x509Certificate3 = certificateFactory.generateCertificate(inputStream) as X509Certificate
+            keyStore.setCertificateEntry("AmazonRootCA1", x509Certificate3)
+            trustManagerFactory.init(keyStore)
+            val sslContext = SSLContext.getInstance("TLS")
```
vasct, line 92: There is a Snyk warning on this line:` Implement secure HTTPS communication. Consider using TLSv1.2 instead`. 

The suggested fix will not do anything useful. 
SSLContext.getInstance("TLS") returns an instance which supports different protocols depending on the Android API version. For api 19-25 this is ["SSLv3", "TLSv1", "TLSv1.1", "TLSv1.2"] and 26+ is ["TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3"]. 
If we change this to SSLContext.getInstance("TLSv1.2") as the fix suggests we will only remove version 1.3 and up from the supported protocols and the older ones will still be available, which defeats the purpose of the warning.

We could 1) create an ssl factory that explicitly supports only TLS 1.2 and 1.3 since the backend support them or 2) disable the warning and keep this as it is. Currently if the backend and device both support TLS 1.2/1.3 they are used

CTLalit: Lets disable the warning for now

darshanclevertap: @vasct @CTLalit I will check how we can disable/bypass this for the time being

darshanclevertap: Todor is going to assist us with bypassing this. I will inform @vasct once we have this done

vasct:
Closing this PR to open a non-wip one.

PR #542, opened 2024-01-23
Task/sdk 3236/migrate to work manager

Anush-Shand:
https://wizrocket.atlassian.net/browse/SDK-3236


clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
```diff
@@ -518,70 +478,84 @@ public Void call() {
 
     @SuppressLint("MissingPermission")
     @RequiresApi(api = VERSION_CODES.LOLLIPOP)
-    private void createOrResetJobScheduler(Context context) {
-
+    private void stopJobScheduler(Context context) {
         int existingJobId = StorageHelper.getInt(context, Constants.PF_JOB_ID, -1);
-        JobScheduler jobScheduler = (JobScheduler) context.getSystemService(JOB_SCHEDULER_SERVICE);
+        if (existingJobId >= 0) {
```
CTLalit, line 483: != 1

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
```diff
@@ -518,70 +478,84 @@ public Void call() {
...
-            int jobid = config.getAccountId().hashCode();
-            JobInfo.Builder builder = new JobInfo.Builder(jobid, componentName);
-            builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);
-            builder.setRequiresCharging(false);
-
-            builder.setPeriodic(pingFrequency * Constants.ONE_MIN_IN_MILLIS, 5 * Constants.ONE_MIN_IN_MILLIS);
-            builder.setRequiresBatteryNotLow(true);
-
-            if (Utils.hasPermission(context, "android.permission.RECEIVE_BOOT_COMPLETED")) {
-                builder.setPersisted(true);
```
CTLalit, line 575: is this related to phone restart and is this handled?

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
```diff
@@ -518,70 +478,84 @@ public Void call() {
...
+
+                // Todo - Earlier flex interval was 5 minutes but it was always defaulted to higher time? Possibly redundant to specify
+                PeriodicWorkRequest request = new PeriodicWorkRequest.Builder(CTPushAmpWorker.class, pingFrequency,
+                        TimeUnit.MINUTES)
+                        .setConstraints(constraints)
+                        .build();
+
+                String workName = existingWorkName.equals("") ? config.getAccountId() : existingWorkName;
+
+                workManager.enqueueUniquePeriodicWork(workName, ExistingPeriodicWorkPolicy.REPLACE, request);
```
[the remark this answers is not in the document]
darshanclevertap: @CTLalit Yes, to support multiple instances. Multi-instance is where different instances of CleverTapAPI are invoked on the same app to send date to different accounts.

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
CTLalit, line 1: why not convert to kotlin?

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/amp/CTPushAmpWorker.kt
```diff
@@ -0,0 +1,16 @@
...
+import androidx.work.Worker
+import androidx.work.WorkerParameters
+import com.clevertap.android.sdk.CleverTapAPI
+import com.clevertap.android.sdk.Logger
+
+class CTPushAmpWorker(context: Context, params: WorkerParameters) : Worker(context, params) {
+
+    override fun doWork(): Result {
+        Logger.v("PushAmpWorker is awake")
+        CleverTapAPI.runJobWork(applicationContext)
```
CTLalit, line 13: do we need asynctask to start in this?

CTLalit (commented):
Check comments.

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
Anush-Shand: WorkManager automatically handles this on it's own
https://stackoverflow.com/questions/50510559/androidx-workmanager-how-to-make-a-scheduled-job-persistent

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
Anush-Shand: Done

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
```diff
@@ -518,70 +478,84 @@ public Void call() {
...
+            return;
         }
 
-        if (needsCreate) {
-            int jobid = config.getAccountId().hashCode();
-            JobInfo.Builder builder = new JobInfo.Builder(jobid, componentName);
-            builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);
-            builder.setRequiresCharging(false);
-
-            builder.setPeriodic(pingFrequency * Constants.ONE_MIN_IN_MILLIS, 5 * Constants.ONE_MIN_IN_MILLIS);
```
Anush-Shand, line 571: @darshanclevertap @piyush-kukadiya We have 5 minutes specified as the flex interval for the job. Any specific reason and do we need the same for work manager?

darshanclevertap: @Anush-Shand `pingFrequency` comes from the server under the key `pf` which is set as 240. So in production, the actual value is 240 mins or 4 hours. 5 mins is the fallback option in case the server does not send it.

Anush-Shand: @darshanclevertap pingfrequency is defaulted to 4 hours in the sdk as well. The second parameter (currently 5 minutes) specifies the flex interval i.e the window in the period (4 hours) in which the Job/Work can be completed
https://developer.android.com/develop/background-work/background-tasks/persistent/getting-started/define-work#flexible_run_intervals

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/amp/CTPushAmpWorker.kt
Anush-Shand: This is not required as the worker is already running on a different thread. The db read is synchronised and the network requests are by default made on postAsyncSafely
Changing the same

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
darshanclevertap: @Anush-Shand Ah, yes. So `setPeriodic` has interval and flex. If you do not provide flex, it will run anytime within the given interval time. We wanted it to be close to every 4 hours, so we set the flex as 5 mins. Hence the Job will run every 4 hours in the last 5 mins. This will make it truly periodic

Anush-Shand: @darshanclevertap 
Got it, we can add the same for work manager. But it reduces the chances of the work actually being completed for that period since there is only a window of 5 minutes to match the constraints

darshanclevertap: @Anush-Shand I think we can do with that. Since it will work in the next run even if one interval is missed.

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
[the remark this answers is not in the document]
Anush-Shand: This has been handled. There is one periodic work request per app. The worker when invoked calls runPushAmpWork() for every available instance

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
```diff
@@ -443,145 +441,120 @@ private boolean alreadyHaveToken(String newToken, PushType pushType) {
...
+        int minute = now.get(Calendar.MINUTE);
 
-                if (lastTS == 0 || lastTS > System.currentTimeMillis() - 24 * 60 * 60 * 1000) {
-                    try {
-                        JSONObject eventObject = new JSONObject();
-                        eventObject.put("bk", 1);
-                        analyticsManager.sendPingEvent(eventObject);
+        Date currentTime = parseTimeToDate(hour + ":" + minute);
+        Date startTime = parseTimeToDate(Constants.DND_START);
+        Date endTime = parseTimeToDate(Constants.DND_STOP);
```
CTLalit, line 458: We should not create new date object everytime.

CTLalit (commented):
Comments added

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
Anush-Shand: As in the startTime and the endTime can be Date constants itself rather than having DND_START and DND_STOP?

CTLalit: we cannot have them as fixed constants i guess because we have to compute them for daily timestamps of DND hours. But the SDF object can be reused, you can explore if there are better ways to do this

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
Anush-Shand: Can be done. The file is quite extensive though

CTLalit: you can take a call for this and resolve.

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/amp/CTPushAmpWorker.kt
Anush-Shand: @piyush-kukadiya Can you please check this?
The Worker itself runs on a separate thread and hence postAsync is not required. 
The db read in runJobWork() is synchronised and the network requests are later by default made on postAsyncSafely

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
Anush-Shand: Sure. I'll pick this up if I have the bandwidth before the release

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
Anush-Shand: This has been addressed

CTLalit (approved):

PR #547, opened 2024-01-31
Task/sdk 3639/refactor geofence unit tests

Anush-Shand:
Java 17 doesn't support powermock. Hence geofence tests have been refactored

### DIFF
--- a/clevertap-geofence/build.gradle
+++ b/clevertap-geofence/build.gradle
@@ -25,7 +25,7 @@ dependencies {
     testImplementation (libs.test.junit.platform.runner)
     testImplementation (libs.test.junit.jupiter.api)
 
-    testImplementation (libs.test.mockito.core)
+    testApi (libs.bundles.mockito)
     testImplementation (libs.test.robolectric)
 
     testImplementation (libs.test.core)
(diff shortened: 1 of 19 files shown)


clevertap-geofence/build.gradle
```diff
@@ -25,15 +25,14 @@ dependencies {
     testImplementation (libs.test.junit.platform.runner)
     testImplementation (libs.test.junit.jupiter.api)
 
-    testImplementation (libs.test.mockito.core)
+    testApi (libs.bundles.mockito)
```
CTLalit, line 28: why did we change to api? the client will have to add it then? or is it pulling from core lib?

CTLalit (commented):
Only one comment for mockstatic, why did we not use the kotlin lib version from MockK?

clevertap-geofence/build.gradle
Anush-Shand: Few unit tests were failing if testImplementation was used, now resolved

Anush-Shand:
@CTLalit mockk wasn't used as then a lot of already working tests would have also needed refactoring. To make it uniform decided to go forward with Mockito
We should use mockk going forward

CTLalit (approved):

PR #548, opened 2024-02-05
Lp/knock/anr jio

CTLalit:
The register method of SDK is synchronised. This might theoretically cause ANR but seems not the cause for client. We have removed the syncronised keyword so main or any thread is not blocked.

Code snippet used to test
`val scope = CoroutineScope(Dispatchers.Default)
        GlobalScope.launch(Dispatchers.Default) {
            val lis = mutableListOf<Deferred<Unit>>()

            (1..1000).map { c ->
                lis.add(
                    scope.async {
                        ActivityLifecycleCallback.register(this@MyApplication)
                        registerActivityLifecycleCallbacks(this@MyApplication)
                    }
                )
            }
            lis.awaitAll()
        }`


darshanclevertap:
@CTLalit The reason we had kept this synchronized was because this class creates an instance of CleverTapAPI and we wanted only one instance to be created and used. Only question is will using `static` give us the same result?

PR #551, opened 2024-02-14
Feat: denied callback for push primer

CTLalit:
- added callback in case of cancel clicked in settings alert
- https://wizrocket.atlassian.net/browse/SDK-3670


clevertap-core/src/main/java/com/clevertap/android/sdk/PushPermissionManager.java
```diff
@@ -99,6 +99,7 @@ public void showFallbackAlertDialog() {
             return Unit.INSTANCE;
         }, () -> {
             if (activity instanceof InAppNotificationActivity) {
+                ((InAppNotificationActivity) activity).notifyPermissionDenied();
```
Anush-Shand, line 102: Can `pushPermissionResultCallback` be passed to `showFallbackAlertDialog()` and hence we don't need a new function in the `InAppNotificationActivity.java`

CTLalit: extending from the current design of passing function lambdas, that is the pattern used in other cases as well.

Anush-Shand: Yess. Understood, I was thinking something like this

    ```
public void showFallbackAlertDialog(InAppNotificationActivity.PushPermissionResultCallback pushPermissionResultCallback) {
        AlertDialogPromptForSettings.show(activity, () -> {
            Utils.navigateToAndroidSettingsForNotifications(activity);
            isFromNotificationSettingsActivity = true;
            return Unit.INSTANCE;
        }, () -> {
            if (activity instanceof InAppNotificationActivity) {
                ((InAppNotificationActivity) activity).didDismiss(null);
                pushPermissionResultCallback.onPushPermissionDeny();
            }
            return Unit.INSTANCE;
        });
    }
```

piyush-kukadiya: @CTLalit @Anush-Shand Does `onResume()` of `InAppNotificationActivity` gets called when dialog is cancelled? If yes then we can reuse this LOC  `isFromNotificationSettingsActivity = true;` which is in `onAccept` of dialog. In that case  `onResume()` of `InAppNotificationActivity` handles notifying deny call but there will be extra if checks in `onResume()` also `isFromNotificationSettingsActivity` naming need to changed.

piyush-kukadiya (commented):
Added comments

clevertap-core/src/main/java/com/clevertap/android/sdk/PushPermissionManager.java
CTLalit: onResume() of InAppNotificationActivity does not get called in this flow as we discussed in the call

piyush-kukadiya (approved):
LGTM

PR #557, opened 2024-02-16
task(SDK-3620) - Updates AGP to 8.2.2

Anush-Shand:
Updates libs.versions.toml

- agp version was wrong because of erroneous conflict resolve
- kotlin plugin version downgraded because of incompatibility

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -3,7 +3,7 @@
 android_compileSdk = "34"
 android_gradle_plugin = "8.2.2"
 android_minSdk = "19"
-kotlin_plugin = "1.7.20"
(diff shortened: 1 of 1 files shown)


CTLalit (approved):

gradle/libs.versions.toml
```diff
@@ -1,9 +1,9 @@
 [versions]
 # Project
 android_compileSdk = "34"
-android_gradle_plugin = "7.4.2"
+android_gradle_plugin = "8.2.2"
 android_minSdk = "19"
-kotlin_plugin = "1.9.0"
+kotlin_plugin = "1.7.20"
```
CTLalit, line 6: need context on this change, can we connect?

PR #581, opened 2024-03-06
feat(MC-1457): Create custom in-app templates definitions

vasct:


vasct:
Example usage would look like this:

Kotlin
```
CleverTapApi.register {
    templatesSet(
        function(isVisual = false) {
            name("function")
            stringArgument("string", "")
            booleanArgument("bool", true)
        },
        template {
            name("template")
            mapArgument("map", mapOf("int" to 5))
        },
    )
}
```
or

```
TemplatesManager.register {
    setOf(
        FunctionBuilder(isVisual = false)
            .name("function")
            .stringArgument("string", "")
            .booleanArgument("bool", true)
            .build(),
        TemplateBuilder()
            .name("template")
            .mapArgument("map", mapOf("int" to 5))
            .build()
    )
}
```

Java:
```
CleverTapApi.register(ctConfig -> {
    Set<CustomTemplate> templates = new HashSet<>();
    templates.add(
            new FunctionBuilder(false)
                    .name("function")
                    .stringArgument("string", "")
                    .booleanArgument("bool", true)
                    .build());

    Map<String, Object> mapArgument = new HashMap<>();
    mapArgument.put("int", 5);
    templates.add(
            new TemplateBuilder()
                    .name("template")
                    .mapArgument("map", mapArgument)
                    .build()
    );
    return templates;
});
```
or

```
CleverTapApi.register(ctConfig ->
        templatesSet(
                new TemplateBuilder()
                        .name("name")
                        .booleanArgument("bool", true)
                        .fileArgument("file")
                        .build(),
                new FunctionBuilder(true)
                        .name("asd")
                        .build()
        )
);
```

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /**
+     * Builder for [CustomTemplate] functions. Function arguments' names cannot contain "."
+     * @param isVisual See [CustomTemplate.isVisual]
+     */
+    class FunctionBuilder(isVisual: Boolean) :
+        Builder<FunctionBuilder>(FUNCTION, allowHierarchicalNames = false, isVisual) {
+
+        override val thisRef: FunctionBuilder = this
```
CTLalit, line 52: why not initialise in const itself?

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
```diff
@@ -0,0 +1,271 @@
...
+            return this
+        }
+
+        fun actionArgument(name: String): TemplateBuilder {
+            addArgument(name, ACTION, null)
+            return this
+        }
+    }
+
+    sealed class Builder<T : Builder<T>>(
```
CTLalit, line 119: how do we handle thrown exceptions? will this cause crash in client or we will catch it internally?

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
```diff
@@ -0,0 +1,271 @@
...
+                    is Byte -> byteArgument(argName, argValue)
+                    is Short -> shortArgument(argName, argValue)
+                    is Int -> intArgument(argName, argValue)
+                    is Long -> longArgument(argName, argValue)
+                    is Float -> floatArgument(argName, argValue)
+                    is Double -> doubleArgument(argName, argValue)
+                    is Boolean -> booleanArgument(argName, argValue)
+                    is String -> stringArgument(argName, argValue)
+                    is Map<*, *> -> mapArgument(argName, argValue as Map<String, Any>)
+                    else -> throw CustomTemplateException("Unsupported value type ${argValue.javaClass} for argument $argName")
```
CTLalit, line 107: consider having config for failing silently.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
```diff
@@ -0,0 +1,271 @@
...
+
+    sealed class Builder<T : Builder<T>>(
+        private val type: CustomTemplateType,
+        private val allowHierarchicalNames: Boolean,
+        private val isVisual: Boolean
+    ) {
+
+        protected abstract val thisRef: T
+
+        private var templateName: String? = null
```
CTLalit, line 127: can we have a val in constructor itself? it will save mutability and force client to assign a name

clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplateTest.kt
```diff
@@ -0,0 +1,226 @@
...
+    @Test
+    fun `builder should throw when name is blank or empty`() {
+        assertThrows<CustomTemplateException> {
+            template {
+                name("")
+            }
+        }
+        assertThrows<CustomTemplateException> {
+            template {
+                name("  ")
```
CTLalit, line 40: how many space characters do we consider?

clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/customtemplates/TemplatesManagerTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.clevertap.android.sdk.CleverTapInstanceConfig
+import com.clevertap.android.shared.test.BaseTestCase
+import org.junit.Test
+import org.junit.jupiter.api.*
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class TemplatesManagerTest : BaseTestCase() {
+
+    override fun setUp() {
```
CTLalit, line 12: there should be before and after block ideally, is there a need to use BaseTestCase()?

CTLalit (commented):
check comments.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
vasct: `this` ref cannot be passed in the super constructor invocation, because it is not initialized at that point.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
vasct: We've designed the API around defining/registering templates to fail fast. We decided on that with these arguments in mind: 1. Template definitions with invalid state (names, argNames, types) are useless - they would either be rejected from the backend or not work as intended when trying to use them later. 2. Template definitions are somewhat declaratively defined and raising an exception would immediately signal for the bad definition and the place it occurs instead of having default values or ignoring bad values (which could result in unexpected behavior for the user). 3. The templates are supposed to be registered around app launch and we assume that developers would run their app at least once when defining templates. Also the templates should be synched with the backend in "dev mode" and that is at least one run. 

Once the templates are successfully registered we would not throw exceptions in order not to crash the clients apps as is with the other APIs in the SDK.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
vasct: Same considerations as above

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
vasct: `templateName` should be validated. Validating in the constructor would have to either have a default value, which would hide the problematic declaration or throw in the constructor which is a bad design. This is one of the main reasons for us to use a builder in this case.

clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/customtemplates/TemplatesManagerTest.kt
vasct: It is extended mainly for having CleverTapInstanceConfig and appCtx provided.

vasct:
Added a few more test cases.

clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/customtemplates/TemplatesManagerTest.kt
vasct: I have moved clearRegisteredProducers()  to an @After method.

CTLalit: We can mock that class for this test case? are we accessing or operating on it?

CTLalit (approved):

clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/customtemplates/TemplatesManagerTest.kt
vasct: It can be mocked. I've changed it in the latest commit.

nzagorchev (approved):

PR #583, opened 2024-03-11
feat(MC-1459): Implement templates sync

vasct:


clevertap-core/src/main/java/com/clevertap/android/sdk/network/NetworkManager.java
```diff
@@ -629,28 +655,39 @@ private boolean handleVariablesResponse(@NonNull Response response) {
...
             return true;
         } else {
-            handleVariablesResponseError(response);
+            handleVarsOrTemplatesResponseError(response, "Variables");
             return false;
         }
     }
 
-    private void handleVariablesResponseError(Response response) {
+    private void handleVarsOrTemplatesResponseError(Response response, String logTag) {
```
CTLalit, line 663: this is used only for logging, is there some wip workflow to add here?

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/DefineTemplatesRequestBody.kt
```diff
@@ -0,0 +1,44 @@
...
+        put(templates.toJSON())
+    }
+
+    override fun toString(): String = jsonArray.toString()
+}
+
+private fun Collection<CustomTemplate>.toJSON(): JSONObject {
+    val templates = this
+
+    return JSONObject().apply {
```
CTLalit, line 21: lint might warn for constants.

CTLalit (approved):
Approved with some small comments.

clevertap-core/src/main/java/com/clevertap/android/sdk/network/NetworkManager.java
vasct: Nothing more is planned for now. In case of error, the backend will not apply any of the changes and we just log the error. This is the same behavior as with the /defineVars endpoint.

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/DefineTemplatesRequestBody.kt
vasct: It does not complain on my IDE. I think directly using the values in this case makes the whole block more readable, but if we must use constants for those cases I can change it.

PR #585, opened 2024-03-28
Task/sdk 3740

CTLalit:
Sample all Real time uninstall tracking support

- adds dependency and also supports real time uninstall tracking as mentioned here https://developer.clevertap.com/docs/uninstall-tracking#real-time-uninstall-tracking-android


Anush-Shand (approved):
LGTM

PR #586, opened 2024-03-29
Fragment activity dependency

CTLalit:
Feat: Remove fragment activity dependency

- client needs to use appcompat activity to host fragments
- routing through our own activity so client has no restrictions
- code cleanup and redundant overides removed
- tested with footer and it seems there are no problems, need to test other cases


CTLalit:
The idea is to have background interactive along with Header/Footer.

Approach:
- moved all fragments to InAppActvity -> problem is interactive background elements
- What we can do? -> attach fragment view to root and not use fragment manager as we do not need back controls/handling.

PR #595, opened 2024-04-10
task(MC-1458): Add custom code templates data in CTInAppNotification

vasct:


clevertap-core/src/main/java/com/clevertap/android/sdk/Constants.java
```diff
@@ -122,6 +122,8 @@ public interface Constants {
     String INAPP_MAX_PER_DAY_KEY = "imp";
     String INAPP_WINDOW = "w";
     String INAPP_KEY = "inApp";
+    String INAPP_TEMPLATE_NAME = "templateName";
```
CTLalit, line 125: can we add new constants in new feature specific package

clevertap-core/src/main/java/com/clevertap/android/sdk/utils/JsonUtils.kt
```diff
@@ -1,9 +1,39 @@
...
+    }
+}
+
+fun JSONObject.getObjectOrNull(name: String): JSONObject? {
+    return try {
+        getJSONObject(name)
+    } catch (je: JSONException) {
+        null
+    }
+}
```
CTLalit, line 27: these methods might be present in JSONObject itself.

CTLalit (approved):
Approved with small comments.

clevertap-core/src/main/java/com/clevertap/android/sdk/utils/JsonUtils.kt
vasct: I will remove getObjectOrNull since there is JSONObject.optJSONObject which behaves the same.  JSONObject.optString() however returns an empty string if the key was not found instead of null and I think getStringOrNull is useful.

clevertap-core/src/main/java/com/clevertap/android/sdk/Constants.java
vasct: I will move them to the sole class that needs them

PR #601, opened 2024-04-19
task(MC-1460): Implement template presenting

vasct:
Add TemplatePresenter to the template definitions
Add TemplateContext API for handling template invocations. Merge template args overrides with default values and provide access to the values. Integrate template presentation in the current in-app display flow

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/CoreState.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/CoreState.java
@@ -7,6 +7,7 @@ import com.clevertap.android.sdk.events.BaseEventQueueManager;
 import com.clevertap.android.sdk.events.EventMediator;
 import com.clevertap.android.sdk.inapp.ImpressionManager;
 import com.clevertap.android.sdk.inapp.InAppController;
+import com.clevertap.android.sdk.inapp.customtemplates.TemplatesManager;
 import com.clevertap.android.sdk.inapp.evaluation.EvaluationManager;
 import com.clevertap.android.sdk.inapp.store.preference.StoreRegistry;
 import com.clevertap.android.sdk.login.LoginController;
@@ -73,6 +74,8 @@ public class CoreState extends CleverTapState {
     private CryptHandler cryptHandler;
 
     private StoreRegistry storeRegistry;
+
+    private TemplatesManager templatesManager;
 
     public CTVariables getCTVariables() {
         return ctVariables;
@@ -311,6 +314,14 @@ public class CoreState extends CleverTapState {
 
     public void setCryptHandler(final CryptHandler cryptHandler) {
         this.cryptHandler = cryptHandler;
+    }
+
+    public TemplatesManager getTemplatesManager() {
+        return templatesManager;
+    }
+
+    public void setTemplatesManager(final TemplatesManager templatesManager) {
+        this.templatesManager = templatesManager;
     }
 
     /**
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
@@ -36,6 +36,8 @@ import com.clevertap.android.sdk.ManifestInfo;
 import com.clevertap.android.sdk.PushPermissionResponseListener;
 import com.clevertap.android.sdk.StorageHelper;
 import com.clevertap.android.sdk.Utils;
+import com.clevertap.android.sdk.inapp.customtemplates.CustomTemplateInAppData;
+import com.clevertap.android.sdk.inapp.customtemplates.TemplatesManager;
 import com.clevertap.android.sdk.inapp.data.InAppResponseAdapter;
 import com.clevertap.android.sdk.inapp.evaluation.EvaluationManager;
 import com.clevertap.android.sdk.inapp.evaluation.LimitAdapter;
@@ -44,6 +46,7 @@ import com.clevertap.android.sdk.network.NetworkManager;
 import com.clevertap.android.sdk.task.CTExecutorFactory;
 import com.clevertap.android.sdk.task.MainLooperHandler;
 import com.clevertap.android.sdk.task.Task;
+import com.clevertap.android.sdk.utils.JsonUtilsKt;
 import com.clevertap.android.sdk.variables.JsonUtil;
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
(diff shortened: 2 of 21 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/CoreState.java
```diff
@@ -313,6 +316,14 @@ public void setCryptHandler(final CryptHandler cryptHandler) {
         this.cryptHandler = cryptHandler;
     }
 
+    public TemplatesManager getTemplatesManager() {
+        return templatesManager;
+    }
+
+    public void setTemplatesManager(final TemplatesManager templatesManager) {
```
CTLalit, line 323: is this setter needed or we always pass it in constructor?

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
```diff
@@ -835,4 +846,24 @@ public void onAppLaunchServerSideInAppsResponse(@NonNull JSONArray appLaunchServ
             addInAppNotificationsToQueue(serverSideInAppsToDisplay);
         }
     }
+
+    private void presentTemplate(final CTInAppNotification inAppNotification) {
+        templatesManager.presentTemplate(inAppNotification, this);
+    }
+
+    private JSONArray filterNonRegisteredCustomTemplates(JSONArray inAppNotifications) {
+        JSONArray filteredInApps = new JSONArray();
```
CTLalit, line 855: can create a generic function .filter for JsonArray

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
```diff
@@ -18,6 +18,11 @@ class CustomTemplate private constructor(
      */
     val name: String,
 
+    /**
+     * The presenter associated with this template.
+     */
+    val presenter: CustomTemplatePresenter<*>,
```
CTLalit, line 24: <*> can we avoid the * wildcard here?

CTLalit, line 24: CustomTemplate seems like a data class which contains only properties for template data. Can we separate data and presenters and not add them together in an object?

clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/evaluation/EvaluationManagerTest.kt
```diff
@@ -40,7 +42,9 @@ class EvaluationManagerTest : BaseTestCase() {
         triggersManager = mockk(relaxed = true)
         limitsMatcher = mockk(relaxed = true)
         storeRegistry = mockk(relaxed = true)
-        evaluationManager = spyk(EvaluationManager(triggersMatcher, triggersManager, limitsMatcher, storeRegistry))
+        templatesManager = mockk(relaxed = true)
+        evaluationManager =
+            spyk(EvaluationManager(triggersMatcher, triggersManager, limitsMatcher, storeRegistry, templatesManager))
```
CTLalit, line 47: names to call args.

CTLalit, line 47: can we remove code smells which are focring us to use spyk() function?

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EvaluationManager.kt
```diff
@@ -42,11 +44,12 @@ import java.util.Locale
  * @property storeRegistry An instance of [StoreRegistry] to access storage for in-app notifications.
  */
 @RestrictTo(LIBRARY)
-class EvaluationManager constructor(
+internal class EvaluationManager(
```
CTLalit, line 47: we can remove @RestrictTo(LIBRARY)

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/TemplatesManager.kt
```diff
@@ -42,4 +47,46 @@ internal class TemplatesManager(templates: Collection<CustomTemplate>) {
...
+
+    fun presentTemplate(notification: CTInAppNotification, inAppListener: InAppListener) {
+        val context = createContextFromInApp(notification, inAppListener) ?: return
+        val template = customTemplates[context.templateName]
+        if (template == null) {
+            logger.info("[CustomTemplates] Cannot find template with name ${context.templateName}")
+            return
+        }
+
+        when (val presenter = template.presenter) {
```
CTLalit, line 59: can we capture presenter type and context in generics rather than using frequent `is` checks?

CTLalit (commented):
Some comments added.

clevertap-core/src/main/java/com/clevertap/android/sdk/CoreState.java
vasct: The constructor of CoreState is only passed a Context. All other fields are set in CleverTapFactory::getCoreState with setters, the same is done with the TemplatesManager.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
vasct: We can avoid it if we introduce a generic type to CustomTemplate. However that would only move the wildcard usage in the classes that use CustomTemplate regardless of the presenter type.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/TemplatesManager.kt
vasct: Presenter and Context are generic for CustomTemplate, but we store and treat CustomTemplates regardless of their inner type (function or template, and the actual types of presenter and context) in all other places. The type checks are needed only when using the presenters, which is not pretty, but should only happen in the TemplateManager.

clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/evaluation/EvaluationManagerTest.kt
vasct: A good idea, but let's keep this outside of this review and do it separately.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/CustomTemplate.kt
vasct: We don't serialize this class and have overridden equals and hashcode. Making it a data class and moving the presenters to an outside list somewhere would introduce complexity, that I think we can avoid for now.

CTLalit (approved):

PR #603, opened 2024-04-23
task(SDK-3795) - Fixes NPE

Anush-Shand:
- Adds null check before string comparison

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
@@ -706,7 +706,7 @@ public class InAppController implements CTInAppNotification.CTInAppNotificationL
             return;
         }
 
-        boolean isHtmlType = inAppNotification.getType().equals(Constants.KEY_CUSTOM_HTML);
+        boolean isHtmlType = inAppNotification.getType() != null && inAppNotification.getType().equals(Constants.KEY_CUSTOM_HTML);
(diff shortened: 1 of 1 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
```diff
@@ -706,7 +706,7 @@ private static void showInApp(Context context, final CTInAppNotification inAppNo
             return;
         }
 
-        boolean isHtmlType = inAppNotification.getType().equals(Constants.KEY_CUSTOM_HTML);
+        boolean isHtmlType = inAppNotification.getType() != null && inAppNotification.getType().equals(Constants.KEY_CUSTOM_HTML);
```
CTLalit, line 709: inAppNotification.getType() -> extract once

CTLalit (commented):
Check comments.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
Anush-Shand: Resolved. Just curious if it only improves readability or does it improve performance as well?

CTLalit: just from a thread safety pov, had seen some problem in other parts of code.

CTLalit (approved):

PR #609, opened 2024-05-21
Add define method for custom in-app templates CleverTapAPI

vasct:


clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
```diff
@@ -1005,6 +1005,10 @@ private static CleverTapAPI fromBundle(final Context context, final Bundle extra
         return fromAccountId(context, _accountId);
     }
 
+    private static boolean isDevelopmentMode(Context context) {
+        return 0 != (context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE);
```
CTLalit, line 1009: make sure null checks are not needed here.

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
```diff
@@ -1089,6 +1093,47 @@ public CustomTemplateContext getActiveContextForTemplate(@NonNull String templat
...
+            return;
+        }
+
+        if (coreState == null) {
+            getConfigLogger().debug("CustomTemplates",
+                    "coreState is null, templates will cannot be synced");
+            return;
+        }
+
+        if (!(coreState.getNetworkManager() instanceof NetworkManager)) {
```
CTLalit, line 1115: we can either get rid of forced abstraction of base class or create proper feature specific abstractions and interface checks.

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
```diff
@@ -1089,6 +1093,47 @@ public CustomTemplateContext getActiveContextForTemplate(@NonNull String templat
...
+            return;
+        }
+
+        if (coreState.getTemplatesManager() == null) {
+            getConfigLogger().debug("CustomTemplates",
+                    "templateManager is null, templates will cannot be synced");
+            return;
+        }
+
+        getCleverTapID(x -> {
```
CTLalit, line 1127: the listeners are not stored in a list. Potentially we might end up missing out on some other listener.
`public void syncVariables()` also has this method call. 
Commenting so you can investigate about this.

CTLalit (commented):
Comments added

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
vasct: Good call, the current implementation of getCleverTapID could have led to problems. I have changed the listeners to be a list.

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
vasct: Added the new method to the base interface

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
vasct: I have added null checks to the invoking method, although it should probably not be needed. This is the context that is provided when creating the CleverTapAPI instance

CTLalit (approved):
lgtm

clevertap-core/src/main/java/com/clevertap/android/sdk/CallbackManager.java
```diff
@@ -49,7 +52,7 @@ public class CallbackManager extends BaseCallbackManager {
 
     private NotificationRenderedListener notificationRenderedListener;
 
-    private OnInitCleverTapIDListener onInitCleverTapIDListener;
+    private final List<OnInitCleverTapIDListener> onInitCleverTapIDListeners = new ArrayList<>();
```
CTLalit, line 55: should be a set and also synchronised. Lets add todo

PR #611, opened 2024-05-23
Chore: Migrate and check changes

CTLalit:
- directly migrate exoplayer -> media3
- take a note of changes
- app is compilable and running
- did not test if features are working fine (in the end we will test)

- NOTE: JUST FOR OBSERVATION, DO NOT MERGE


clevertap-core/src/main/java/com/clevertap/android/sdk/Utils.java
```diff
@@ -439,7 +439,7 @@ private static boolean checkForExoPlayer() {
         try {
             className = Class.forName("com.google.android.exoplayer2.ExoPlayer");
             className = Class.forName("com.google.android.exoplayer2.source.hls.HlsMediaSource");
-            className = Class.forName("com.google.android.exoplayer2.ui.StyledPlayerView");
+            className = Class.forName("com.google.android.exoplayer2.ui.PlayerView");
```
CTLalit, line 442: this is incorrect code and we will change it.

PR #614, opened 2024-05-27
task(SDK-3806) - Adds fix for incorrectly stored data in userProfiles…

Anush-Shand:
- Fixes the issue during DB migration
- Removes the "$D_" prefix from date related property values
- Removes the "$set" and "$add" key from incorrectly stored multi-valued properties

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/db/CtDatabase.kt
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/db/CtDatabase.kt
@@ -6,6 +6,9 @@ import android.database.sqlite.SQLiteDatabase
 import android.database.sqlite.SQLiteOpenHelper
 import com.clevertap.android.sdk.CleverTapInstanceConfig
 import com.clevertap.android.sdk.Constants
+import com.clevertap.android.sdk.Constants.COMMAND_ADD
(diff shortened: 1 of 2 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/db/CtDatabase.kt
```diff
@@ -96,13 +101,63 @@ class DatabaseHelper internal constructor(val context: Context, val config: Clev
...
+     * Removes the "$set" and "$add" key from incorrectly stored multi-valued properties
+     */
+    private fun migrateDataString(dataString: String): String {
+        return try {
+            val jsonObject = JSONObject(dataString)
+            val keys = jsonObject.keys()
+
+            while (keys.hasNext()) {
+                val key = keys.next()
+                var value = jsonObject.get(key)
```
CTLalit, line 138: check if this throws exception or will cause NPE further in func

CTLalit (approved):

clevertap-core/src/main/java/com/clevertap/android/sdk/db/CtDatabase.kt
Anush-Shand: if `value is String` and `value is JSoNObject ` will return false if value is null and mitigate any NPE

PR #618, opened 2024-06-03
task(SDK-3852) - Refactors to remove syncWithUpstream unused feature

Anush-Shand:
- Removes syncWithUpstream and related functions in LocalDataStore.java
- Simplifies setting and removing functions in LocalDataStore.java
- Removes response class and it's associated callers from NetworkManager

The key "evpr" is never received in any response used for syncing with BE. It is an unused feature and hence can be removed


CTLalit (approved):
Approved, did not check logic for LocalDataStore change. We will discuss that in main PR merge.

PR #620, opened 2024-06-06
task(SDK-3853) - Fixes Stopped/Deleted campaign is not cleared in the SDK in ClientSide mode

Anush-Shand:
https://wizrocket.atlassian.net/browse/SDK-3853

Fix 
- Even when an empty JSoN array is received in response for `inapp_notifs_cs`, save it 
- Same is done for `inapp_notifs_ss`  for clarity even though it will not cause any issue previously

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/CTXtensions.kt
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/CTXtensions.kt
@@ -216,9 +216,22 @@ inline fun <reified T> JSONArray.iterator(foreach: (element: T) -> Unit) {
  *         The first value of the Pair is `true` if the JSONArray is not null and not empty, `false` otherwise.
  *         The second value of the Pair is the non-empty JSONArray if it exists, or `null` otherwise.
  */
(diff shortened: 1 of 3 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/CTXtensions.kt
```diff
@@ -216,11 +216,24 @@ inline fun <reified T> JSONArray.iterator(foreach: (element: T) -> Unit) {
...
+ * @param key The key to retrieve the JSONArray.
+ * @return A [Pair] indicating success and the retrieved JSONArray.
+ *         The first value of the Pair is `true` if the JSONArray is not null, `false` otherwise.
+ *         The second value of the Pair is the non-null JSONArray if it exists, or `null` otherwise.
+ */
+fun JSONObject.safeGetJSONArray(key: String): Pair<Boolean, JSONArray?> {
+    val list: JSONArray = optJSONArray(key) ?: return Pair(false, null)
+    return Pair(list.length() >= 0, list.takeIf { it.length() >= 0 })
+}
+
```
CTLalit, line 236: both the functions look similar

PR #629, opened 2024-06-27
Task/mc 1790/headers inapps eval incorrectly attached t

Anush-Shand:
- Separates the `eval` and `suppressed` lists for profile and normal events and now attaches based on the type of queue


clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EvaluationManager.kt
```diff
@@ -51,13 +67,13 @@ internal class EvaluationManager(
     private val templatesManager: TemplatesManager
 ) : NetworkHeadersListener {
 
-    // Internal list to track server-side evaluated campaign IDs.
+    // Internal map to track server-side evaluated campaign IDs. This map is used to identify the evaluatedIDs for raised and profile events individually
     @VisibleForTesting
-    internal var evaluatedServerSideCampaignIds: MutableList<Long> = ArrayList()
+    internal var evaluatedServerSideCampaignIds: MutableMap<String, MutableList<Long>> = mutableMapOf(Constants.RAISED to mutableListOf(), Constants.PROFILE to mutableListOf())
```
CTLalit, line 72: Since we changing this from list to map is migration required as we use Shared prefs to load this.

Anush-Shand: MIgration handled in `InAppStore.kt`
@CTLalit

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/store/preference/InAppStore.kt
```diff
@@ -167,25 +169,41 @@ class InAppStore(
...
+        if (evaluatedServerSideInAppIds.isNullOrBlank()) return JSONObject()
+
+        return try {
+            // Try to convert the string to a JSONObject
+            JSONObject(evaluatedServerSideInAppIds)
+        } catch (jsonException: JSONException) {
+            // If it fails, convert the string to a JSONArray
+            val jsonArray = JSONArray(evaluatedServerSideInAppIds)
+            // Wrap the JSONArray in a JSONObject
+            JSONObject().put(Constants.RAISED, jsonArray)
```
CTLalit, line 185: so basically we expect the catch block to trigger first time and it correcting the data? 
is there no code in migration where are check version upgrade.

Anush-Shand: Yess. That was the idea
In such a case, version check won't be needed

CTLalit (commented):
Lets make this with readable methods even if we dont change the logic of migration in try catch.
move catch block code to method named `migrateToct700` or similiar so that we know in future this was done.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/store/preference/InAppStore.kt
Anush-Shand: @CTLalit addressed in latest commit

CTLalit (approved):
lgtm

PR #630, opened 2024-07-02
Task/sdk 3898/app launched repeated

Anush-Shand:
- Start postAsyncSafely earlier to prevent race conditions in case of custom clevertap-id
- Remove lock as postAsyncSafely and onUserLogin will now always run one after the other
- Race condition was due to `onUserLogin` running on the client's thread and `asyncProfileSwitchUser` on the executor thread. The deviceId was not updated yet and hence asyncProfileSwitchUser was called once from getDefaultInstance and again when the user logged in immediately


piyush-kukadiya (approved):
Test out execution sequence on postAsyncSafely. Rest looks good.

PR #635, opened 2024-07-05
task(SDK - 3915) - Fixes dismiss callback when inapp was rotated

Anush-Shand:
- Reverts to 5.2.2 and handles deprecated onBackPressed in the recommended way
- This also fixes alert template related crashes when rotated

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/InAppNotificationActivity.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/InAppNotificationActivity.java
@@ -17,6 +17,7 @@ import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
 import android.os.Bundle;
 import android.view.WindowManager;
+import androidx.activity.OnBackPressedCallback;
 import androidx.annotation.NonNull;
 import androidx.core.content.ContextCompat;
 import androidx.fragment.app.FragmentActivity;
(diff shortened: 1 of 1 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/InAppNotificationActivity.java
```diff
@@ -64,6 +61,14 @@ public interface PushPermissionResultCallback {
 
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        getOnBackPressedDispatcher().addCallback(this, new OnBackPressedCallback(true) {
+            @Override
+            public void handleOnBackPressed() {
+                finish();
+                didDismiss(null);
+            }
+        });
```
CTLalit, line 70: do we not have to remove callback?

CTLalit (commented):
Check comments.

clevertap-core/src/main/java/com/clevertap/android/sdk/InAppNotificationActivity.java
Anush-Shand: @CTLalit This is the documentation

```
Receive callbacks to a new OnBackPressedCallback when the given LifecycleOwner is at least started.
This will automatically call addCallback(OnBackPressedCallback) and remove the callback as the lifecycle state changes. As a corollary, if your lifecycle is already at least started, calling this method will result in an immediate call to addCallback(OnBackPressedCallback).
When the LifecycleOwner is destroyed, it will automatically be removed from the list of callbacks. The only time you would need to manually call OnBackPressedCallback.remove() is if you'd like to remove the callback prior to destruction of the associated lifecycle.
If the Lifecycle is already destroyed when this method is called, the callback will not be added.
```

CTLalit (approved):
lgtm

PR #649, opened 2024-07-18
Fix bugs encountered while testing edge cases SDK-3934

piyush-kukadiya:
1. When an unresponsive image URL was encountered, a null pointer exception was thrown while accessing the bitmap. This exception caused the rest of the URL downloads to be skipped because the coroutines caught the exception and subsequently skipped the remaining downloads.


CTLalit (changes requested):
Let us use exception handlers for coroutines to handle elegantly.

PR #653, opened 2024-07-25
Add registering of custom templates through a json definition

vasct:


clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/JsonTemplatesProducer.kt
```diff
@@ -0,0 +1,252 @@
...
+ * @param templatesPresenter A presenter for all templates in the json definitions. Required if there
+ * is at least one template with type "template".
+ * @param functionsPresenter A presenter for all functions in the json definitions. Required if there
+ * is at least one template with type "function".
+ */
+open class JsonTemplatesProducer(
+    private val jsonTemplatesDefinition: String,
+    private val templatesPresenter: TemplatePresenter?,
+    private val functionsPresenter: FunctionPresenter?
+) : TemplateProducer {
```
CTLalit, line 63: @Throws(CustomTemplateException::class) -> for those who use this in java

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/JsonTemplatesProducer.kt
```diff
@@ -0,0 +1,252 @@
...
+
+    override fun defineTemplates(ctConfig: CleverTapInstanceConfig): Set<CustomTemplate> {
```
CTLalit, line 65: Seems like `ctConfig: CleverTapInstanceConfig` is unused

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/JsonTemplatesProducer.kt
```diff
@@ -0,0 +1,252 @@
...
+                templatesSet.add(
+                    createTemplateFromJson(
+                        templateName,
+                        jsonDefinitions.getJSONObject(templateName)
+                    )
+                )
+            }
+            templatesSet
+        } catch (je: JSONException) {
+            throw CustomTemplateException("Invalid JSON format for templates' definitions", je)
```
CTLalit, line 80: can we also point to the place where there is a problem so client can fix, lets do it if feasible.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/JsonTemplatesProducer.kt
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+    }
+
+    private fun addJsonArgumentsToBuilder(
+        builder: CustomTemplate.FunctionBuilder,
+        json: JSONObject
+    ) {
+        for (key in json.keys()) {
+            val argumentJson = json.getJSONObject(key)
```
CTLalit, line 173: might throw exception, make sure it is handled (wont throw ever/catch it)

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/DefineTemplatesRequestBody.kt
```diff
@@ -23,15 +23,15 @@ private fun Collection<CustomTemplate>.toJSON(): JSONObject {
         putObject("definitions") {
             for (template in templates) {
                 putObject(template.name) {
-                    put("type", template.type.stringName)
+                    put("type", template.type.toString())
```
CTLalit, line 26: hope this change is intentional.

CTLalit (commented):
Added some comments.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/JsonTemplatesProducer.kt
vasct: CustomTemplatesException should not be a checked exception, as it is not designed to be caught. I've added comments to all methods that throw to make the exceptions more clear.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/JsonTemplatesProducer.kt
vasct: I have let `JsonTemplateProducer` be an open class. If clients need to use `ctConfig` to differentiate between instances, they could extend this class, override `defineTemplates` and use `ctConfig` to check whether to call the super implementation or not.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/JsonTemplatesProducer.kt
vasct: The `cause` exception will contain additional information. In case of JSONException it points to the character number where the exception occurred, which should help in figuring out the problem.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/customtemplates/JsonTemplatesProducer.kt
vasct: It is caught in the try-catch block of `defineTemplates` and re-thrown with a `CustomTemplateException`. I've added comments to all methods that throw to make that more clear.

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/DefineTemplatesRequestBody.kt
vasct: Yes, I've changed `toString()` to return `stringName`.

Anush-Shand (approved):

CTLalit (approved):

PR #670, opened 2024-09-24
Network layer changes for handshake support.

CTLalit:
Jira : https://wizrocket.atlassian.net/browse/SDK-4042


clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/CtApi.kt
```diff
@@ -147,28 +140,18 @@ internal class CtApi(
...
-            if (cachedSpikyDomain.isNotNullAndBlank()) {
-                return cachedSpikyDomain!!
-            }
-        } else {
-            if (cachedDomain.isNotNullAndBlank()) {
-                return cachedDomain!!
-            }
+        val toCheckCached = if (isViewedEvent) { cachedSpikyDomain } else { cachedDomain }
+        if (toCheckCached.isNotNullAndBlank()) {
+            return toCheckCached!!
```
vasct, line 154: Just a suggestion: We can drop the redundant !! mark if we define ```isNotNullAndBlank``` with a contract:
```kotlin
@OptIn(ExperimentalContracts::class)
fun String?.isNotNullAndBlank() : Boolean {
    contract { returns(true) implies (this@isNotNullAndBlank != null) }
    return isNullOrBlank().not()
}
```
It is still an experimental feature, but I think it is safe for this use case. It just tells the compiler that the function includes a null check.

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/CtApi.kt
```diff
@@ -184,27 +167,20 @@ internal class CtApi(
...
-            if (spikyProxyDomain.isNotNullAndBlank()) {
-                return false
-            }
-        } else {
-            if (proxyDomain.isNotNullAndBlank()) {
-                return false
-            }
+        val toCheckProxy = if (isViewedEvent) { spikyProxyDomain } else { proxyDomain }
+        if (toCheckProxy.isNotNullAndBlank()) {
+            clearCachedHandshakeDomain.invoke()
```
vasct, line 172: This lambda is non-null, so you can directly call it: ```clearCachedHandshakeDomain()```.

clevertap-core/src/test/java/com/clevertap/android/sdk/network/api/CtApiTest.kt
```diff
@@ -242,15 +242,29 @@ class CtApiTest {
             customHandshakeDomain = null
         }
 
+        var counter = 0
```
vasct, line 245: You can replace the counter logic with mockk' verify methods:

```kotlin
        val clearCachedHandshakeDomain = mockk<() -> Unit>(relaxed = true)

        // assert
        assertEquals(
            expected = false,
            actual = ctApi.needsHandshake(false, clearCachedHandshakeDomain)
        )
        verify(exactly = 1) { clearCachedHandshakeDomain() }

        assertEquals(
            expected = false,
            actual = ctApi.needsHandshake(true, clearCachedHandshakeDomain)
        )
        verify(exactly = 2) { clearCachedHandshakeDomain() }
```

Anush-Shand (approved):

clevertap-core/src/test/java/com/clevertap/android/sdk/network/api/CtApiTest.kt
CTLalit: the logic is removed in the chained pr as we no longer need a lambda, thanks for the hints, will remember in next set of UTs.

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/CtApi.kt
CTLalit: done.

clevertap-core/src/main/java/com/clevertap/android/sdk/network/api/CtApi.kt
CTLalit: removed this bit in chained PR as it is no longer needed.

clevertap-core/src/main/java/com/clevertap/android/sdk/CoreState.java
```diff
@@ -242,13 +242,11 @@ public void setLoginController(final LoginController loginController) {
         this.loginController = loginController;
     }
 
-    @Override
-    public BaseNetworkManager getNetworkManager() {
+    public NetworkManager getNetworkManager() {
```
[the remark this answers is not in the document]
CTLalit: removed the not necessary abstraction so there are no references to that class and hence no casts.

PR #672, opened 2024-09-27
feat(SDK-4076): Changes need handshake logic

CTLalit:
- needs no handshake when mentioned first time, migration to be done via redirection headers in /a1 api call
- changes methods in ctapi, also need to clear any previously cached handshake performance info (did/did not do)
Jira : https://wizrocket.atlassian.net/browse/SDK-4076


Anush-Shand (approved):

piyush-kukadiya (commented):
GTM

PR #681, opened 2024-10-29
bug(MC-2299): Fix c2a param not tracked for open_url and close actions

vasct:
This fixes tracking callToAction if it exists as a parameter to the links opened from the in-apps webview. There is also a frontend fix directly in the in-app HTML, that would use triggerAction instead of opening links.

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppBaseFragment.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppBaseFragment.java
@@ -91,7 +91,7 @@ public abstract class CTInAppBaseFragment extends Fragment {
 
             String callToAction = formData.getString(Constants.KEY_C2A);
             if (callToAction != null) {
-                final String[] parts = callToAction.split("__dl__");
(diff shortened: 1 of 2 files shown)


CTLalit (approved):

Anush-Shand (approved):

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppBaseFragment.java
```diff
@@ -89,13 +89,13 @@ void openActionUrl(String url) {
         try {
             final Bundle formData = UriHelper.getAllKeyValuePairs(url, false);
 
-            String actionParts = formData.getString(Constants.KEY_C2A);
-            String callToAction = null;
-            if (actionParts != null) {
-                final String[] parts = actionParts.split("__dl__");
+            String callToAction = formData.getString(Constants.KEY_C2A);
+            if (callToAction != null) {
+                final String[] parts = callToAction.split("__dl__");
```
Anush-Shand, line 94: __dl__ Can be extracted as a constant?

PR #682, opened 2024-10-30
task(SDK-4057) - Revamped Timer Notification Template

Anush-Shand:
https://wizrocket.atlassian.net/browse/SDK-4057

- The idea is such that the timer notification shown will be the compulsory FGS notification
- The existing implementation has been used to prepare the notification builder for the timer template
- If the service is not declared in the manifest, then the older approach is used to render the timer notification


clevertap-core/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
     <!-- Required runtime permission to display notifications on Android 13 -->
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE_REMOTE_MESSAGING"/>
```
CTLalit, line 6: we need to ask clients to put this permission as it might have some compliances on playstore.

darshanclevertap: I agree with @CTLalit - this should be mentioned in the docs and needs to be optional. All the code related to FGS should be with a permission check with all cases handled properly

Anush-Shand: Done
Most SDK's are using these permissions and they are not as critical as like an AlarmManager,  it was merged by default in the final manifest

Moving it to the app to be safe

Anush-Shand:
@darshanclevertap @piyush-kukadiya @CTLalit 
Another caveat here is that Android enforces that a FGS must call `startForeground` within 5s https://stackoverflow.com/questions/44425584/context-startforegroundservice-did-not-then-call-service-startforeground/72754189#72754189

- Image download can take time for a notification and hence this might lead to a crash

Options
- Call `startForeground` with a placeholder notification and update/dismiss it after notification builder for timer template has been generated
- Timebound notification builder for timer template to 4s. If it fails, skip the notification, call startForeground() and stopSelf()

darshanclevertap:
@Anush-Shand - My suggestion is to go with option 1 - Call `startForeground` with a placeholder notification and update/dismiss it after the notification builder for the timer template has been generated.

Placeholders can have the title & message while the image & timer loads. Once the image loads, we have the notification ID to update the notification. If you could share a demo video of this on Slack, it would help.

Option 2 is risky because if some delay happens more than 4s, the app will crash. See the 2nd comment here -> https://stackoverflow.com/a/45047542/8301966

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
```diff
@@ -3379,6 +3381,28 @@ public void renderPushNotificationOnCallerThread(@NonNull INotificationRenderer
 
     }
 
+    @RestrictTo(Scope.LIBRARY_GROUP)
+    public NotificationCompat.Builder getPushNotificationOnCallerThread(@NonNull INotificationRenderer iNotificationRenderer, Context context,
+                                                                Bundle extras) {
+        CleverTapInstanceConfig config = coreState.getConfig();
+        try {
+            synchronized (coreState.getPushProviders().getPushRenderingLock()) {
```
CTLalit, line 3389: avoid locking if there is only one push provider available?

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapAPI.java
```diff
@@ -3379,6 +3381,28 @@ public void renderPushNotificationOnCallerThread(@NonNull INotificationRenderer
...
+                config.getLogger().verbose(config.getAccountId(),
+                        "returning push on caller thread with id = " + Thread.currentThread().getId());
```
CTLalit, line 3391: maybe print if its main thread.

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
```diff
@@ -148,68 +146,101 @@ private PushProviders(
...
+     * @param notificationId A custom id to build a notification
+     */
+    @RestrictTo(Scope.LIBRARY)
+    public NotificationCompat.Builder _getNotification(final Context context, final Bundle extras, int notificationId) {
+        try {
+            int generatedNotificationId = generateNotificationId(notificationId, extras);
+            NotificationCompat.Builder nb = getPreparedNotificationBuilder(context, extras, generatedNotificationId);
+            if (nb != null) {
+                return nb;
+            }
```
CTLalit, line 179: you are masking the logic behind a nullable NotificationBuilder return method, ideally check from data itself if we should short circuit.

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
```diff
@@ -148,68 +146,101 @@ private PushProviders(
...
-                            .storeUninstallTimestamp();
-                    String pingFreq = extras.getString("pf", "");
-                    if (!TextUtils.isEmpty(pingFreq)) {
-                        updatePingFrequencyIfNeeded(context, Integer.parseInt(pingFreq));
-                    }
-                    return;
+                        .verbose(config.getAccountId(),
+                                "Push notification message is empty, not rendering");
+                baseDatabaseManager.loadDBAdapter(context)
+                        .storeUninstallTimestamp();
```
CTLalit, line 235: this might cause lag on main thread, confirm that is not happening.

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
```diff
@@ -148,68 +146,101 @@ private PushProviders(
...
+                String pingFreq = extras.getString("pf", "");
+                if (!TextUtils.isEmpty(pingFreq)) {
+                    updatePingFrequencyIfNeeded(context, Integer.parseInt(pingFreq));
```
CTLalit, line 238: `Integer.parseInt` can throw, make sure it is handled.

clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/PushTemplateNotificationHandler.java
```diff
@@ -34,23 +36,45 @@ public boolean onActionButtonClick(final Context context, final Bundle extras, f
...
         } catch (Throwable throwable) {
             PTLog.verbose("Error parsing FCM payload", throwable);
         }
         return true;
     }
 
+    public boolean shouldRenderTimerTemplateUsingFGS(Context applicationContext, TemplateType templateType) {
+        return ManifestValidator.isComponentPresentInManifest(applicationContext, "com.clevertap.android.pushtemplates.TimerTemplateService", ManifestValidator.ComponentType.SERVICE)
+                && templateType == TemplateType.TIMER
+                && areRevampedTimerTemplatePermissionsGranted(applicationContext);
```
CTLalit, line 59: `templateType == TemplateType.TIMER` can be first condition so short circuit happens without much computation

clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/TemplateRenderer.kt
```diff
@@ -300,6 +303,12 @@ class TemplateRenderer : INotificationRenderer, AudibleNotification {
                         basicTemplateBundle
                     )
                 }
+                else {
+                    PTLog.verbose("Initial notification not in tray, not Rendering Timer Terminal Notification")
+                }
+
+                val intent = Intent(context, TimerTemplateService::class.java)
+                context.stopService(intent)
             }, (delay - 100).toLong())
```
CTLalit, line 312: lets see if we can solve for not using such a code in this task.

clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/TimerTemplateService.kt
```diff
@@ -0,0 +1,62 @@
...
+    override fun onStartCommand(intent: Intent, flags: Int, startId: Int): Int {
+        val message = intent.extras ?: return super.onStartCommand(intent, flags, startId)
+        val templateRenderer = TemplateRenderer(this@TimerTemplateService, message)
+        PTLog.verbose("Running TimerTemplateService")
+
+        val cleverTapAPI = CleverTapAPI.getGlobalInstance(
+            this@TimerTemplateService,
+            PushNotificationUtil.getAccountIdFromNotificationBundle(message)
+        )
+        val config: CleverTapInstanceConfig? = cleverTapAPI?.coreState?.config
```
CTLalit, line 24: can we run this code inside the task itself?

sample/src/main/AndroidManifest.xml
```diff
@@ -57,13 +59,13 @@
...
-            android:value="YOUR ACCT ID" />
+            android:value="TEST-4RW-944-876Z" />
         <meta-data
             android:name="CLEVERTAP_TOKEN"
-            android:value="YOUR ACCT TOKEN" />
+            android:value="TEST-44c-024" />
         <meta-data
             android:name="CLEVERTAP_REGION"
-            android:value="YOUR ACCT REGION"/>
+            android:value="eu1" />
```
CTLalit, line 68: delete these details

clevertap-pushtemplates/src/main/java/com/clevertap/android/pushtemplates/PushTemplateNotificationHandler.java
Anush-Shand: FIxed in [116f4ca](https://github.com/CleverTap/clevertap-android-sdk/pull/682/commits/116f4ca67a359dedb381a763fcc75c199f12bb13)

clevertap-core/src/main/java/com/clevertap/android/sdk/pushnotification/PushProviders.java
Anush-Shand: Yess. This is legacy code. It only has been refactored into a separate function

Anush-Shand:
Closing this in favour of https://github.com/CleverTap/clevertap-android-sdk/pull/749

PR #697, opened 2024-11-25
feat(SDK-4183): Event properties normalisation

CTLalit:
- backend sends normalised event properties, we need to compare on sdk side based on normalised event which client app might fire

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/variables/JsonUtil.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/variables/JsonUtil.java
@@ -2,6 +2,8 @@ package com.clevertap.android.sdk.variables;
 
 import android.text.Editable;
 import com.clevertap.android.sdk.Logger;
+import com.clevertap.android.sdk.Utils;
+
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -50,6 +52,10 @@ public class JsonUtil {
   }
 
   public static <T> Map<String, T> mapFromJson(JSONObject object) {
+    return mapFromJson(object, false);
+  }
+
+  public static <T> Map<String, T> mapFromJson(JSONObject object, boolean normaliseNames) {
     if (object == null) {
       return null;
     }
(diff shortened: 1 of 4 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/variables/JsonUtil.java
```diff
@@ -68,7 +74,9 @@ public static <T> Map<String, T> mapFromJson(JSONObject object) {
         value = null;
       }
       T castedValue = uncheckedCast(value);
-      result.put(key, castedValue);
+      if (normaliseNames) {
+        result.put(Utils.getNormalizedName(key), castedValue);
+      }
```
vasct, line 79: Missing the else statement for when normaliseNames=false

clevertap-core/src/main/java/com/clevertap/android/sdk/variables/JsonUtil.java
```diff
@@ -50,6 +52,10 @@ public static JSONArray listToJsonArray(Iterable<?> list) throws JSONException {
   }
 
   public static <T> Map<String, T> mapFromJson(JSONObject object) {
+    return mapFromJson(object, false);
+  }
+
+  public static <T> Map<String, T> mapFromJson(JSONObject object, boolean normaliseNames) {
     if (object == null) {
       return null;
     }
```
vasct, line 61: keysIterator is actually of type `Iterator<String>`. If it is changed the cast of `key` to (String) will be redundant.

vasct (changes requested):

clevertap-core/src/main/java/com/clevertap/android/sdk/variables/JsonUtil.java
CTLalit: done.

vasct (dismissed):

vasct:
@CTLalit check the failing tests.

vasct:
Also it could be an easier implementation if we only normalize the names when comparing the properties instead of always getting them normalized. This is what I did for the event names and profile properties.

PR #700, opened 2024-11-28
feat(SDK-4183): changes normalisation to follow LP logic

CTLalit:
- makes changes to event names and properties and also to charged event logic
- tries exact name match first
- tries match with normalised event name
- tries match after normalising both event name/property and triggers we get from BE

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapter.kt
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapter.kt
@@ -34,7 +34,7 @@ class EventAdapter(
     val profileAttrName: String? = null // for profile events only
 ) {
 
-    private val systemPropToKey = mapOf(
+    internal val systemPropToKey = mapOf(
         "CT App Version" to CLTAP_APP_VERSION,
         "ct_app_version" to CLTAP_APP_VERSION,
         "CT Latitude" to CLTAP_LATITUDE,
--- a/clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapterTest.kt
+++ b/clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapterTest.kt
@@ -30,6 +30,49 @@ class EventAdapterTest : BaseTestCase() {
     }
 
     @Test
+    fun testGetPropertyPresentSpecialCases() {
+        // Arrange
+        val eventProperties = mapOf(
+            "name" to "John",
+            Constants.CLTAP_PROP_CAMPAIGN_ID to "cpid",
+            Constants.NOTIFICATION_ID_TAG to "some_wzrk_id",
+            Constants.CLTAP_PROP_VARIANT to "some variant",
+            Constants.INAPP_WZRK_PIVOT to "some_wzrk_pivot",
+        )
+        val eventAdapter = EventAdapter("eventName", eventProperties)
+
+        val triggerProps = listOf(
+            Constants.CLTAP_PROP_CAMPAIGN_ID,
+            Constants.NOTIFICATION_ID_TAG,
+            Constants.CLTAP_PROP_VARIANT,
+            Constants.INAPP_WZRK_PIVOT
+        )
+
+        val expectedResults = listOf(
+            TriggerValue("cpid"),
+            TriggerValue("some_wzrk_id"),
+            TriggerValue("some variant"),
+            TriggerValue("some_wzrk_pivot")
+        )
+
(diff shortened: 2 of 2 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapter.kt
```diff
@@ -81,11 +81,20 @@ class EventAdapter(
         return items
             .filterNotNull()
             .map { productMap: Map<String, Any> ->
-                val normalisedMap = productMap.map {
-                    Utils.getNormalizedName(it.key) to it.value
-                }.toMap()
 
-                TriggerValue(normalisedMap[Utils.getNormalizedName(propertyName)])
+                var op = productMap[propertyName]
```
Anush-Shand, line 85: What is the fullform of op?

Anush-Shand (approved):

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapter.kt
piyush-kukadiya, line 85: Can't we replace logic L85 to L95 with `evaluateActualPropertyValue()`. Seems duplicate code.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapter.kt
```diff
@@ -110,22 +119,35 @@ class EventAdapter(
...
         }
 
         return value
     }
+
+    private fun evaluateActualPropertyValue(propertyName: String): Any? {
+        var value = eventProperties[propertyName]
+
+        if (value == null) {
+            value = eventProperties[Utils.getNormalizedName(propertyName)]
```
piyush-kukadiya, line 142: Can we maintain and use a HashMap of `originalEventName` to `normalizeEventName`?  Calling `Utils.getNormalizedName()` again and again seems not useful.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapter.kt
```diff
@@ -110,22 +119,35 @@ class EventAdapter(
...
+                CLTAP_PROP_VARIANT -> evaluateActualPropertyValue(INAPP_WZRK_PIVOT)
+                INAPP_WZRK_PIVOT -> evaluateActualPropertyValue(CLTAP_PROP_VARIANT)
+                else -> systemPropToKey[propertyName]?.let { evaluateActualPropertyValue(it) }
             }
         }
 
         return value
     }
+
+    private fun evaluateActualPropertyValue(propertyName: String): Any? {
```
piyush-kukadiya, line 138: Calling `getActualPropertyValue("propABC")` multiple times will re-compute same value multiple times. If possible can be memoized.

piyush-kukadiya (commented):
Added comments

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapter.kt
CTLalit: it is just a local var to capture trigger value returned for the key.

CTLalit: this bit is for the items from charged event product, the method you have mentioned is for event parameters (event properties).

piyush-kukadiya: Screenshot 2024-11-29 at 12 12 00 PM
As per screenshot we can see duplicate code can be replaced by single method. Specific Map can be passed to method.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapter.kt
CTLalit: this was done in closed pr, the calls are already optimised in the matching callers. Hence these will not over-run and we can create comparision map on the fly.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/evaluation/EventAdapter.kt
CTLalit: it will be called for different `getActualPropertyValue `
The collision is a special case/really rare occurence.

piyush-kukadiya (approved):
LGTM

PR #706, opened 2024-12-04
Perform event/profile property normalization before querying DB MC-2432

piyush-kukadiya:
- [x] Add `normalizedEventName` column to `userEventLogs` table,
- [x] When inserting event log to `userEventLogs` table, also include `normalizedEventName`
- [x] When querying `userEventLogs` table, query by `normalizedEventName`
- [x] Convert actual event name to normalized before interaction with  `userEventLogs` table
- [x] When returning `UserEventLog`, also include `normalizedEventName` field in the class
- [x] Fix and modify broken tests
- [x] Add missing `firstTime` profile property check to `TriggersMatcher.matchFirstTimeOnly()`
- [x] Add tests for `firstTime` profile property check from `TriggersMatcher.matchFirstTimeOnly()`
- [x] Bump gradle and gradle plugin versions

Screenshot 2024-11-28 at 1 23 53 AM


gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 # Project
 android_compileSdk = "34"
-android_gradle_plugin = "8.2.2"
+android_gradle_plugin = "8.7.0"
```
Anush-Shand, line 4: Does upgrading the AGP need any other changes regarding deprecated features etc.

clevertap-core/src/main/java/com/clevertap/android/sdk/db/CtDatabase.kt
```diff
@@ -235,12 +236,12 @@ private val CREATE_EVENTS_TABLE = """
 private val CREATE_USER_EVENT_LOGS_TABLE = """
     CREATE TABLE ${Table.USER_EVENT_LOGS_TABLE.tableName} (
         ${Column.EVENT_NAME} STRING NOT NULL,
+        ${Column.NORMALIZED_EVENT_NAME} STRING NOT NULL,
```
Anush-Shand, line 239: 1. Do we need both event_name and normalised_event_name? Or can we normalise the event_name as soon the user raises it and use that throughout the code

2. Is it a product requirement for `getUserEventLog` to provide the original event_name as well? If we have both columns in the db, and the client calls
- recordEvent("test sample") 
- then recordEvent("Testsample")
- then getUserEventLog("test sample") 

Here the client might be expecting _countOfEvent as 1 and the eventName as "test sample"_ but will get _countOfEvent as 2 and eventName as Testsample_

Is this expected?

clevertap-core/src/main/java/com/clevertap/android/sdk/LocalDataStore.java
```diff
@@ -53,6 +57,7 @@ public class LocalDataStore {
 
     private final DeviceInfo deviceInfo;
     private final Set<String> userEventLogKeys = Collections.synchronizedSet(new HashSet<>());
+    private final Map<String, String> normalizedEventNames = new HashMap<>();
```
Anush-Shand, line 60: Do we have this to reduce repeated calls to `getNormalizedName`?
Seems like a tradeoff between memory and execution-time

vasct (approved):

gradle/libs.versions.toml
piyush-kukadiya: Nope. Mostly, all releases targets bug fixes and lint warnings.

clevertap-core/src/main/java/com/clevertap/android/sdk/LocalDataStore.java
piyush-kukadiya: >Do we have this to reduce repeated calls to getNormalizedName?

YES. To improve performance. Memoize already computed stuffs.

piyush-kukadiya: Memory will not be an issue here. Memory will be released once App is terminated.

clevertap-core/src/main/java/com/clevertap/android/sdk/db/CtDatabase.kt
Anush-Shand: Okayy.

1. recordEvent("test sample")
2. then recordEvent("Testsample")
3. then getUserEventLog("test sample")

Here the client might be expecting countOfEvent as 1 and the eventName as "test sample" but will get countOfEvent as 2 and eventName as Testsample. The returned data might be slightly inconsistent, if that is acceptable then can merge

piyush-kukadiya: Recording events: `LeVeL PaSsEd`, `Level Passed`, and `levelpassed` all appear as `LeVeL PaSsEd` since this is the way the event was recorded for the first time (This is how currently backend works, if you push events and check user's profile activity on dashboard). So first variant of event is always considered as actual event and when you need counts it will be 3 as all events are same after normalization (removing white space and making smaller case).

Similarly pushing profile property <k:v> `MoViE: 1`, `moviE: 2`, and `movie: 3`  appear as MoViE:

clevertap-core/src/main/java/com/clevertap/android/sdk/LocalDataStore.java
piyush-kukadiya: Common events can occur multiple times in the same session like `Product Viewed`, `Add to cart`, `Product Details Viewed`, `Item increment`, `Item decrement` and many. Also it's not just events, we have huge payload of CS in-apps in which case same event name can appear multiple times. CS in-apps can have 100 in-apps, each in-app can have multiple triggers, in that way I believe it's better to memoize. Also If we put `log()` in `Utils.getNormalizedName()` we can see it get's call multiple times for already computed value.

CTLalit (approved):

CTLalit:
Why is AGP upgraded in this PR? not a part of this.

Anush-Shand (approved):

PR #709, opened 2024-12-20
Multi-triggers updates

vasct:
Move update of LocalStore to EventQueueManager.processEvent Remove getUserEventLogFirstTs and getUserEventLogLastTs. Replace usage with getUserEventLog
Change getUserEventLogCount to return 0 when no events match

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/db/DBManager.kt
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/db/DBManager.kt
@@ -17,9 +17,12 @@ internal class DBManager(
     private val ctLockManager: CTLockManager
 ) : BaseDatabaseManager {
 
+    private companion object {
+        private const val USER_EVENT_LOG_ROWS_PER_USER = 2_048 + 256 // events + profile props
+        private const val USER_EVENT_LOG_ROWS_THRESHOLD = 5 * USER_EVENT_LOG_ROWS_PER_USER
+    }
+
     private var dbAdapter: DBAdapter? = null
-    private val userEventLogsThreshold = 11_520 //12.59 MB table and index size
-    private val numberOfRowsToCleanupForUserEventLogs = 2_304 //( 2048 events + 256 profile props =  1 user)
 
     @WorkerThread
     @Synchronized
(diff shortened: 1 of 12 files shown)


CTLalit (approved):

clevertap-core/src/main/java/com/clevertap/android/sdk/db/DBManager.kt
```diff
@@ -33,7 +36,7 @@ internal class DBManager(
             dbAdapter.cleanupStaleEvents(PUSH_NOTIFICATION_VIEWED)
             dbAdapter.cleanUpPushNotifications()
             dbAdapter.userEventLogDAO()
-                .cleanUpExtraEvents(userEventLogsThreshold, numberOfRowsToCleanupForUserEventLogs)
+                .cleanUpExtraEvents(USER_EVENT_LOG_ROWS_THRESHOLD, USER_EVENT_LOG_ROWS_THRESHOLD)
```
nzagorchev, line 39: This seems wrong to me, shouldn't it be:
```
.cleanUpExtraEvents(USER_EVENT_LOG_ROWS_THRESHOLD, USER_EVENT_LOG_ROWS_PER_USER)
```

clevertap-core/src/main/java/com/clevertap/android/sdk/usereventlogs/UserEventLogDAOImpl.kt
nzagorchev, line 1: The removed methods are those that were used by `CleverTapAPI` and are no longer needed, correct?
The other change is removing the `defaultValueExtractor` so we can return 0 if there are no records and -1 if there is an error in getting the int from the cursor or exception in the whole block. Is my understanding correct?

nzagorchev (approved):

clevertap-core/src/main/java/com/clevertap/android/sdk/usereventlogs/UserEventLogDAOImpl.kt
vasct: Yes, the removed methods will no longer be used, since I removed them from the CleverTapAPI. The other change is removing the whole readEventColumnByDeviceIdAndNormalizedEventName method, since it was now only used in one place. I move the logic in readEventCountByDeviceIdAndNormalizedEventName and changed the default and error cases integer.

PR #710, opened 2024-12-23
bug(MC-2455): Fix file variables update to a null value

vasct:
Fix VarCacheTest tests


CTLalit (commented):
Code changes LGTM

But i want to understand if we want to trigger file is ready callback in such an event?

CTLalit (approved):

PR #752, opened 2025-02-14
fix(SDK-4385): Handle cases when OnInitCleverTapIDListener might be null

vasct:


clevertap-core/src/main/java/com/clevertap/android/sdk/CallbackManager.java
```diff
@@ -282,8 +282,10 @@ public void removeOnInitCleverTapIDListener(@NonNull final OnInitCleverTapIDList
     @Override
     public void notifyCleverTapIDChanged(final String id) {
         Handler mainHandler = new Handler(Looper.getMainLooper());
-        for (OnInitCleverTapIDListener listener : onInitCleverTapIDListeners) {
-            mainHandler.post(() -> listener.onInitCleverTapID(id));
+        for (final OnInitCleverTapIDListener listener : onInitCleverTapIDListeners) {
+            if (listener != null) {
```
piyush-kukadiya, line 286: @vasct Can we list down all possible rare cases? Is this due to multi-threading? If yes, is it possible to use synchronized list?

vasct: The onInitCleverTapIDListeners list can contain null values, supplied for example through `cleverTapApi.getCleverTapID(listener);` with listener that is `null. We don't set listeners to null anywhere in our code, so I suspect that a race condition could happen in the client code leading to the addition of a null listener. I could also add the null check in the API method, but it is already marked as `NonNull` and I would have to also suppress the lint warning for the redundant null check.

brijesh-kumar-sharma-zomato:
Please get this merged, we are getting some crashes in production due to this. @vasct @piyush-kukadiya

piyush-kukadiya (approved):

Anush-Shand (approved):

brijesh-kumar-sharma-zomato:
@vasct @CTLalit @piyush-kukadiya by when we are aiming to release the fix. We need to bump the version to fix the crash in our app. TIA

brijesh-kumar-sharma-zomato:
@vasct @CTLalit @piyush-kukadiya by when we will release this fix in next clever tap version?

Issue #673, opened 2024-10-09
StrictMode policy violation occurred while displaying InApp notification


boskicaruso:
**Describe the bug**
StrictMode policy violation occurred while displaying InApp notification

**Screenshots/Logs**
StrictMode policy violation; ~duration=12 ms: android.os.strictmode.DiskReadViolation
	at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1679)
	at libcore.io.BlockGuardOs.access(BlockGuardOs.java:74)
	at libcore.io.ForwardingOs.access(ForwardingOs.java:128)
	at android.app.ActivityThread$AndroidOs.access(ActivityThread.java:8551)
	at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:313)
	at java.io.File.exists(File.java:813)
	at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:818)
	at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:809)
	at android.app.ContextImpl.getPreferencesDir(ContextImpl.java:765)
	at android.app.ContextImpl.getSharedPreferencesPath(ContextImpl.java:990)
	at android.app.ContextImpl.getSharedPreferences(ContextImpl.java:605)
	at android.content.ContextWrapper.getSharedPreferences(ContextWrapper.java:223)
	at com.clevertap.android.sdk.store.preference.CTPreference.sharedPrefs$clevertap_core_release(CTPreference.kt:194)
	at com.clevertap.android.sdk.store.preference.CTPreference.readString(CTPreference.kt:27)
	at com.clevertap.android.sdk.inapp.store.preference.ImpressionStore.getLongListFromPrefs(ImpressionStore.kt:74)
	at com.clevertap.android.sdk.inapp.store.preference.ImpressionStore.read(ImpressionStore.kt:29)
	at com.clevertap.android.sdk.inapp.store.preference.ImpressionStore.write(ImpressionStore.kt:39)
	at com.clevertap.android.sdk.inapp.ImpressionManager.recordImpression(ImpressionManager.kt:42)
	at com.clevertap.android.sdk.InAppFCManager.didShow(InAppFCManager.java:115)
	at com.clevertap.android.sdk.inapp.InAppController.inAppNotificationDidShow(InAppController.java:516)
	at com.clevertap.android.sdk.InAppNotificationActivity.didShow(InAppNotificationActivity.java:280)
[...]

PR #753, opened 2025-02-17
fix(git-673) - Moves recording impressions and counts to an io thread

Anush-Shand:
https://github.com/CleverTap/clevertap-android-sdk/issues/673

SDK-4503

After an InApp is displayed, it's impressions are recorded on the main thread. Even though it's simply accessing the shared prefs, this comes up in the StrictModeViolation

Clients are concerned this may be causing ANRs in the worst case. 
There's a possibility that for InApps linked to the AppLaunch event, the prefs might not be loaded into memory yet. This along with slow app startup time might be leading to ANRs

There could be concerns regarding race conditions here?

## Summary by CodeRabbit

- **New Features**
  - Enhanced in-app notifications for a more responsive and consistent user experience.
  - Streamlined the notification delivery process to ensure timely displays while effectively managing notification frequency.

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
@@ -678,31 +678,6 @@ public class InAppController implements InAppListener,
             return;
         }
 
-        final InAppNotificationListener listener = callbackManager.getInAppNotificationListener();
-
-        final boolean goFromListener;
-
-        if (listener != null) {
-            final HashMap<String, Object> kvs;
-
-            if (inAppNotification.getCustomExtras() != null) {
-                kvs = Utils.convertJSONObjectToHashMap(inAppNotification.getCustomExtras());
-            } else {
-                kvs = new HashMap<>();
-            }
-
-            goFromListener = listener.beforeShow(kvs);
-        } else {
-            goFromListener = true;
-        }
-
-        if (!goFromListener) {
-            logger.verbose(config.getAccountId(),
-                    "Application has decided to not show this in-app notification: " + inAppNotification
-                            .getCampaignId());
-            showInAppNotificationIfAny();
-            return;
-        }
         checkLimitsBeforeShowing(context, inAppNotification, config, this);
         incrementLocalInAppCountInPersistentStore(context, inAppNotification);
     }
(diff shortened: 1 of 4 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
```diff
@@ -722,7 +703,7 @@ public void run() {
             showInAppNotificationIfAny();
             return;
         }
-        showInApp(context, inAppNotification, config, this);
+        checkLimitsBeforeShowing(context, inAppNotification, config, this);
```
piyush-kukadiya, line 706: @Anush-Shand We are checking limits after  `beforeShow()` callback is invoked.  `beforeShow()` must be invoked only if we are displaying InApps, there are chances that we call `beforeShow()` but don't display inapp.

piyush-kukadiya (changes requested):
Left one comment, rest looks good

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/InAppController.java
Anush-Shand: Fixed in [e11264f](https://github.com/CleverTap/clevertap-android-sdk/pull/753/commits/e11264f08de7bc64d89af8688799a42a8d31976e)

piyush-kukadiya (approved):

sample/src/main/java/com/clevertap/demo/ui/main/HomeScreenViewModel.kt
```diff
@@ -556,15 +556,15 @@ class HomeScreenViewModel(private val cleverTapAPI: CleverTapAPI?) : ViewModel()
...
             }
 
             "11-1" -> {
-                cleverTapAPI?.pushEvent("Footer InApp without image")
+                cleverTapAPI?.discardInAppNotifications()
             }
 
             "11-2" -> {
-                cleverTapAPI?.pushEvent("Header")
+                cleverTapAPI?.resumeInAppNotifications()
```
CTLalit, line 567: is this intentional?

CTLalit (approved):
Approved with one comment.

sample/src/main/java/com/clevertap/demo/ui/main/HomeScreenViewModel.kt
Anush-Shand: Yes intentional. We didn't have these controls in the sample app. This will help with testing

Pushing event can be done from the events buttons in the sample app

PR #775, opened 2025-03-19
Push prompt system template

vasct:


clevertap-core/src/main/java/com/clevertap/android/sdk/PushPermissionHandler.kt
```diff
@@ -0,0 +1,142 @@
...
+import android.content.Context
+import android.content.pm.PackageManager
+import androidx.core.app.ActivityCompat
+import androidx.core.content.ContextCompat
+import com.clevertap.android.sdk.Constants.NOTIFICATION_PERMISSION_REQUEST_CODE
+import com.clevertap.android.sdk.inapp.AlertDialogPromptForSettings
+import java.lang.ref.WeakReference
+
+internal class PushPermissionHandler(
+    private val config: CleverTapInstanceConfig,
```
CTLalit, line 13: Lets not pass entire config as it reduces testability.
You only need it to use methods of `CTPreferenceCache` which also actually does not require full config.
Might be a good time to reduce this dependencies.

Also consider using `PushPermissionHandler ` in InAppNotificationActivity as it contains similar code (check once)

vasct: I've used it in InAppNotificationActivity as well as in InboxActivity. There was a lot of code duplication between those two before.

clevertap-core/src/test/java/com/clevertap/android/sdk/inapp/InAppActionHandlerTest.kt
```diff
@@ -100,4 +102,10 @@ class InAppActionHandlerTest : BaseTestCase() {
         assertTrue(inAppActionHandler.isPlayStoreReviewLibraryAvailable())
     }
 
+    private fun getMockCtConfig(): CleverTapInstanceConfig {
+        val mockCtConfig = mockk<CleverTapInstanceConfig>(relaxed = true)
+        every { mockCtConfig.logger } returns mockk(relaxed = true)
+        return mockCtConfig
+    }
```
CTLalit, line 109: we are already using a CleverTapInstanceConfig instance from Fixtures class, do take a look if needed.

CTLalit (approved):
LGTM

PR #783, opened 2025-03-27
Remove the dialog when navigating to settings from push permission prompt

vasct:
Remove unused string resources
Remove CTStringResources and CTStringResourcesTest - they are not used
Remove JavaToKtInterOp from the sample package - it is not used


clevertap-core/src/main/java/com/clevertap/android/sdk/Utils.java
```diff
@@ -472,14 +472,10 @@ public static void navigateToAndroidSettingsForNotifications(Context context) {
...
             intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-        } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+        } else {
             intent.setAction("android.settings.APP_NOTIFICATION_SETTINGS");
             intent.putExtra("app_package", context.getPackageName());
             intent.putExtra("app_uid", context.getApplicationInfo().uid);
-        } else {
-            intent.setAction(Settings.ACTION_APPLICATION_DETAILS_SETTINGS);
-            intent.addCategory(Intent.CATEGORY_DEFAULT);
-            intent.setData(Uri.parse("package:" + context.getPackageName()));
```
CTLalit, line 482: else branch seems unreachable but is the code in it required?

PR #794, opened 2025-04-16
task(SDK-4537) - Removes extra space from manual_carousel.xml

Anush-Shand:
- Removes padding from shape and adds margin to layout


CTLalit (approved):
Approved but would be nice if we can have screenshots.

PR #809, opened 2025-05-23
Add invisible FragmentHostActivity to support custom HTML in-app messages on NativeActivity

simonbullen:
This PR adds support for rendering CleverTap’s fragment-based in-app notifications (e.g. custom HTML) in Unreal Engine projects that use a NativeActivity, which does not support fragments.

Problem
=======
CleverTap’s SDK assumes that the current activity is a FragmentActivity, but Unreal-based Android apps typically use NativeActivity. As a result, certain in-app notification types (notably HTML-based ones) fail to display or cause runtime exceptions.

Solution
=======
Introduced a minimal FragmentHostActivity that subclasses AppCompatActivity and provides a safe fragment container.

Added logic to detect when the current activity is not a FragmentActivity, defer the notification, and launch FragmentHostActivity instead.

Once launched, the host activity proceeds with CleverTap’s regular rendering logic.

This approach isolates fragment-specific logic to a dedicated host, without modifying the behavior for apps already using FragmentActivity.

### DIFF
--- a/clevertap-core/src/main/AndroidManifest.xml
+++ b/clevertap-core/src/main/AndroidManifest.xml
@@ -15,6 +15,13 @@
             android:configChanges="keyboardHidden"
             android:theme="@style/Theme.AppCompat.DayNight.DarkActionBar" />
 
+        <activity
+            android:name="com.clevertap.android.sdk.FragmentHostActivity"
+            android:theme="@style/CleverTapFragmentHostTheme"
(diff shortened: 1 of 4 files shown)


CTLalit (changes requested):
@simonbullen I think this is apply changes to all SDK flows. Consider putting behind a flag which can be provided to core-sdk from Unity (any other third party) and make it configurable?

Issue #816, opened 2025-06-03
Restricted class expected to be used in callback


akash-syna:
**Describe the bug**
CTInAppNotification.java is declared as restricted to library. But as part of onShow() method in InAppNotificationListener.java, the client app using the SDK is expected to override onShow() & hence utilise CTInAppNotification instance. This leads to compilation warning & may result in inappropriate/unknown behaviour.

**To Reproduce**
Warning thrown in Android Studio : "CTInAppNotification can only be accessed from within the same library (com.clevertap.android:clevertap-android-sdk)"

**Expected behavior**
The interfaces exposed must not be restricted.

**Screenshots/Logs**

**Environment (please complete the following information):**
 - CleverTap SDK Version v6.0.0, the same is visible in latest code also in github master

**Additional context**

Anush-Shand:
This has been fixed in v7.5.0

PR #817, opened 2025-06-04
Git/816/fixes scope ct inapp notification

Anush-Shand:
https://github.com/CleverTap/clevertap-android-sdk/issues/816

## Summary by CodeRabbit

- **Tests**
  - Migrated geofence module tests from Java/Mockito to Kotlin/MockK, improving maintainability and consistency.
  - Rewrote and enhanced test coverage for geofence components, including API, receivers, factories, adapters, tasks, and utilities.
  - Simplified and consolidated test utility classes and fake data providers in Kotlin.
  - Removed redundant or outdated test files and dependencies.

- **Chores**
  - Updated test dependencies for geofence modules, removing unused libraries and instrumentation test dependencies.

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppNotification.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppNotification.java
@@ -8,9 +8,6 @@ import android.os.Bundle;
 import android.os.Parcel;
 import android.os.Parcelable;
 
-import androidx.annotation.RestrictTo;
-import androidx.annotation.RestrictTo.Scope;
-
 import com.clevertap.android.sdk.Constants;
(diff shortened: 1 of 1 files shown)


vasct:
If we are going to make CTInAppNotification a public API class, we should also make all inapp model classes public: CTInAppNotificationMedia, CTInAppNotificationButton, CTInAppType. Why do we need the CTInAppNotification being public ?

vasct (changes requested):
If we are going to make CTInAppNotification a public API class, we should also make all inapp model classes public: CTInAppNotificationMedia, CTInAppNotificationButton, CTInAppType. Why do we need the CTInAppNotification being public ?

vasct:
@Anush-Shand We should do it in both places. I will add it to the other PR, but no need to block this fix on the other PR being merged

PR #828, opened 2025-06-11
fix(SDK-4752,SDK-4872): Handle fullscreen in-apps

vasct:
Insets margins are no longer applied for in in-apps fragments that are in fullscreen.
Change InAppNotificationActivity landscape fullscreen mode to be edge-to-edge

## Summary by CodeRabbit

- **New Features**
  - Improved fullscreen support for in-app notifications and web views, ensuring system bars are fully hidden in landscape mode.
- **Bug Fixes**
  - Adjusted margin and inset handling for in-app messages to display correctly in both fullscreen and non-fullscreen modes.

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppBaseFullHtmlFragment.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppBaseFullHtmlFragment.java
@@ -1,6 +1,9 @@
 package com.clevertap.android.sdk.inapp;
 
+import static android.view.WindowManager.LayoutParams.FLAG_FULLSCREEN;
+
 import android.annotation.SuppressLint;
+import android.app.Activity;
 import android.content.Context;
 import android.content.res.Configuration;
 import android.graphics.drawable.ColorDrawable;
@@ -8,11 +11,14 @@ import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.Window;
 import android.webkit.WebViewClient;
 import android.widget.RelativeLayout;
 import android.widget.RelativeLayout.LayoutParams;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+
 import com.clevertap.android.sdk.CTWebInterface;
 import com.clevertap.android.sdk.CleverTapAPI;
 import com.clevertap.android.sdk.Constants;
@@ -23,6 +29,7 @@ import com.clevertap.android.sdk.customviews.CloseImageView;
 public abstract class CTInAppBaseFullHtmlFragment extends CTInAppBaseFullFragment {
 
     protected CTInAppWebView webView;
+    protected boolean isFullscreen = false;
 
     @Override
     public void onAttach(@NonNull Context context) {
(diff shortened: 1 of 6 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppBaseFullHtmlFragment.java
```diff
@@ -192,4 +201,14 @@ private void cleanupWebView() {
...
+
+    private void updateFullscreenInfo() {
+        Activity activity = getActivity();
+        if (activity != null) {
+            Window window = activity.getWindow();
+            if (window != null) {
+                isFullscreen = (window.getAttributes().flags & FLAG_FULLSCREEN) != 0;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
vasct: InAppNotificationActivity still sets the FLAG_FULLSCREEN flag

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppBaseFullHtmlFragment.java
```diff
@@ -23,6 +29,7 @@
 public abstract class CTInAppBaseFullHtmlFragment extends CTInAppBaseFullFragment {
 
     protected CTInAppWebView webView;
+    protected boolean isFullscreen = false;
```
CTLalit, line 32: do we need to save this state? We can directly call the `getFullscreenInfo()` method in reDraw method itself? Is it possible to do it inplace?

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppHtmlCoverFragment.java
```diff
@@ -34,7 +35,7 @@ protected RelativeLayout.LayoutParams getLayoutParamsForCloseButton() {
     @Override
     public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, Bundle savedInstanceState) {
         View inAppView = super.onCreateView(inflater, container, savedInstanceState);
-        if (inAppView != null) {
+        if (!isFullscreen && inAppView != null) {
```
CTLalit, line 38: @piyush-kukadiya @vasct can we apply insets at one place in base class `CTInAppBaseFullHtmlFragment` itself?

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppBaseFullHtmlFragment.java
vasct: It is possible. I update it only in the places where it can change to save time during redraw.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppHtmlCoverFragment.java
vasct: The insets are different in the different fragments. It could be possible to extract some common logic, but I am not sure if it would be worth it right now. Maybe after the Kotlin migration of those files.

clevertap-core/src/main/java/com/clevertap/android/sdk/inapp/CTInAppBaseFullHtmlFragment.java
CTLalit: all these methods are synchronous so the total time i believe is not saved at all? correct me if i am wrong.

PR #848, opened 2025-07-02
Removes locking while accessing executors in clevertap.

CTLalit:


CTLalit:
@coderabbitai review

clevertap-core/src/main/java/com/clevertap/android/sdk/task/CTExecutorFactory.java
```diff
@@ -14,21 +12,22 @@ public class CTExecutorFactory {
...
     public static CTExecutors executors(CleverTapInstanceConfig config) {
         if (config == null) {
             throw new IllegalArgumentException("Can't create task for null config");
         }
-        CTExecutors executorForAccount = executorMap.get(config.getAccountId());
+
+        String accountId = config.getAccountId();
+
+        // lock-free read
+        CTExecutors executorForAccount = executorMap.get(accountId);
```
vasct: This looks like a `computeIfAbsent` logic. It is only available in API24+, but we can use it with an API check, so when minVersion exceeds it, it will become a warning and we can remove the old logic:
```java
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
            return executorMap.computeIfAbsent(accountId, (key) -> new CTExecutors(config));
        } else if (executorForAccount == null) {...
```

CTLalit: @vasct done, please check changes

Anush-Shand (approved):

PR #856, opened 2025-07-08
task(SDK-5031) - Local Inapp parity with iOS

Anush-Shand:
We are trying to achieve the following change in behaviour
**Half-Inter**
1. For  Allow button clicked Add actionExtras -> {wzrk_c2a = "Allow", wzrk_id = ""}  to onDismissed callback
2. For  Deny button additional onPushPermissionResponse with accepted -> false

**Alert**
1. For  Allow button clicked Add actionExtras -> {wzrk_c2a = "Allow", wzrk_id = ""}  to onDismissed callback
2. For  Deny button clicked Add actionExtras -> {wzrk_c2a = "Deny", wzrk_id = ""}  to onDismissed callback
3. For  Deny button additional onPushPermissionResponse with accepted -> false

Not raise click event for close button

Backlog -> Dismiss local inapp first and then show the OS prompt like system inapp functions

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/InAppNotificationActivity.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/InAppNotificationActivity.java
@@ -275,7 +275,7 @@ public final class InAppNotificationActivity extends FragmentActivity implements
     @Override
     public void onPushPermissionResult(boolean isGranted) {
         Bundle data = null;
-        if(inAppNotification != null && inAppNotification.isLocalInApp()) {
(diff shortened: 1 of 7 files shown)


clevertap-core/src/main/java/com/clevertap/android/sdk/InAppNotificationActivity.java
```diff
@@ -452,9 +463,13 @@ private void onAlertButtonClickLegacy(final CTInAppNotificationButton button) {
     private void onAlertButtonClick(CTInAppNotificationButton button, boolean isPositive) {
```
CTLalit, line 463: can we simplify by early returns and add test coverage?

PR #860, opened 2025-07-10
Encryption in transit setter for multi instance.

CTLalit:
- allows multi instance to have encryption in transit enabled via config setter.


clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapInstanceConfig.java
```diff
@@ -578,6 +578,10 @@ public int getEncryptionLevel() {
         return encryptionLevel;
     }
 
+    public void setEncryptionInTransit(String encryptionInTransit) {
+        this.encryptionInTransit = encryptionInTransit;
```
piyush-kukadiya: getter returns boolean but setter sets string, can we have same data type?

piyush-kukadiya (approved):
Left comment

clevertap-core/src/main/java/com/clevertap/android/sdk/CleverTapInstanceConfig.java
CTLalit: changed to boolean @piyush-kukadiya

piyush-kukadiya (approved):

PR #1036, opened 2026-07-26
[SDK-5955] App Inbox: repaint the built-in inbox list after pull-to-refresh

piyush-kukadiya:
## Summary

Pull-to-refresh in the built-in App Inbox fetched and stored new messages correctly, but the visible list never updated — the spinner stopped and the stale list stayed on screen. The auto-refresh wiring was removed in v4.6.6 (SDK-2330) because adapter rebinds broke video playback, and was never replaced. This PR implements the UI refresh explicitly and safely, scoped strictly to the user's pull gesture.

## Changes (one commit per step, each independently shippable)

1. **Stale-position hardening** (`0d9c33a`) — the 2s mark-read timer now skips if its holder was rebound to a different message (crash + wrong-message-attribution vector, reachable even before this feature); `didShow` carries the `CTInboxMessage` instead of a bind-time row index; `didClick` ignores out-of-range positions.
2. **`MediaPlayerRecyclerView.prepareForListRebind()`** (`1826668`) — detaches the shared video surface from its holder *before* an adapter rebind. `stop()` alone leaves an orphaned player view inside a recycled row — the exact SDK-2330 regression class.
3. **`CTInboxListViewFragment.refreshList()`** (`2f9e272`) — on pull success: re-queries the committed store, swaps the shared message list **in place** (the adapter aliases it), forward-merges read state (a just-read message never flips back to unread while its async store write is in flight), skips the repaint entirely when id/read/content are identical (a playing video is never interrupted by a no-change pull), and lazily builds the list UI when a tab that opened empty gets its first messages.
4. **Cross-tab consistency** (`ae56590`) — `CTInboxActivity.refreshAllInboxListFragments()` repaints every resident tab from the same snapshot, so a server-deleted message can't stay tappable in a sibling tab. No listener registration — the customer's `CTInboxListener` slot is untouched, and nothing refreshes without an explicit user pull. Dead `inboxContentUpdatedListener` field (unused since v4.6.6) removed.
5. **QA logs** (`fa1de50`, `e40d182`) — verbose-only logs on every decision point (stale-timer skip, no-op guard, read-merge count, video detach, fan-out count, failure path) so manual testing can verify which path ran from logcat; ids and counts only, no message content.
6. **Empty no-tabs inbox fix** (`c113284`) — no-tabs mode with 0 messages never created the fragment, so an empty inbox had no pull surface at all. The activity now always hosts the fragment, which owns the empty state inside the SwipeRefreshLayout (matches the tabs branch, which never gated on count).

## API / dependency impact

- **Zero public API changes** — every touched class is `@RestrictTo(LIBRARY)`.
- **Zero production dependency changes.** `material` added as `testImplementation` only, so tests can inflate `inbox_activity.xml` (TabLayout).
- No ProGuard/consumer-rules impact.

## Deliberately rejected designs

- **DiffUtil** — holders capture bind-time positions; not rebinding "unchanged" holders makes stale-position bugs worse, and in-place `isRead` mutation corrupts content comparison in both directions.
- **Global `inboxMessagesDidUpdate` listener** (the pre-v4.6.6 design) — mark-read fires it per message → refresh storms and a refresh→rebind→mark-read→refresh feedback loop; background fetches would yank the list while the user reads.

## Testing

- 3,424 unit tests green (`clevertap-core:testDebugUnitTest`), including 21 new tests: in-place-swap invariant, read forward-merge, no-op guard matrix, empty↔non-empty transitions, stale-timer identity guard, swipe callback matrix, tab fan-out, and full-activity launch tests for the empty no-tabs case.
- Manual device checklist (video paths are not unit-testable — please gate on this, both ExoPlayer and Media3 flavors): pull with video playing → list updates and video resumes in the best visible row; pull with nothing new → video keeps playing untouched; pull on an empty inbox/tab → messages appear; server-side delete-all → back to the no-message view; rotation mid-fetch; throttled second pull within 5 min → spinner stops, list unchanged.
- Note: `./gradlew detekt` currently fails on JDK 21 (`Invalid value (21) passed to --jvm-target`) — pre-existing on develop, unrelated to this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inbox refresh behavior, including empty-state handling and pull-to-refresh updates.
  * Refreshing inbox content now preserves read status and avoids unnecessary UI updates.
  * Updated multi-tab inboxes so all visible lists can refresh together.
  * Prevented delayed read indicators from affecting messages after list items are rebound.
  * Improved video playback stability while inbox content refreshes.
  * Improved inbox handling for empty lists and invalid item selections.

### DIFF
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
@@ -19,7 +19,6 @@ import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
 import androidx.viewpager.widget.ViewPager;
 
-import com.clevertap.android.sdk.CTInboxListener;
 import com.clevertap.android.sdk.CTInboxStyleConfig;
 import com.clevertap.android.sdk.CleverTapAPI;
 import com.clevertap.android.sdk.CleverTapInstanceConfig;
@@ -65,8 +64,6 @@ public class CTInboxActivity extends FragmentActivity implements CTInboxListView
     private WeakReference<InboxActivityListener> listenerWeakReference;
 
     private CleverTapAPI cleverTapAPI;
-
-    private CTInboxListener inboxContentUpdatedListener = null;
 
     private PushPermissionHandler pushPermissionHandler;
 
@@ -238,6 +235,20 @@ public class CTInboxActivity extends FragmentActivity implements CTInboxListView
         super.onDestroy();
     }
 
+
+    /**
+     * Repaints every resident inbox list fragment from the already-committed message
+     * store, so tabs never show contradicting data after a refresh. Called only from
+     * the pull-to-refresh success path (explicit user action) — never from background
+     * updates, so an on-screen list never changes without a user gesture.
+     */
+    void refreshAllInboxListFragments() {
+        for (Fragment fragment : getSupportFragmentManager().getFragments()) {
+            if (fragment instanceof CTInboxListViewFragment && fragment.isAdded()) {
+                ((CTInboxListViewFragment) fragment).refreshList();
+            }
+        }
+    }
 
     @Override
     public void messageDidClick(Context baseContext, int contentPageIndex, CTInboxMessage inboxMessage, Bundle data,
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
@@ -130,13 +130,10 @@ public class CTInboxActivity extends FragmentActivity implements CTInboxListView
             // otherwise pull-to-refresh has no surface on an empty inbox.
             final FrameLayout listViewFragmentLayout = findViewById(R.id.list_view_fragment);
             listViewFragmentLayout.setVisibility(View.VISIBLE);
-            boolean fragmentExists = false;
-            for (Fragment fragment : getSupportFragmentManager().getFragments()) {
-                if (fragment.getTag() != null && !fragment.getTag().equalsIgnoreCase(getFragmentTag())) {
-                    fragmentExists = true;
-                }
-            }
-            if (!fragmentExists) {
+            // On recreation (rotation, process restore) the FragmentManager restores the
+            // previously added fragment before this code runs — adding again would stack
+            // a duplicate list (and a duplicate video player) into the same container.
+            if (getSupportFragmentManager().findFragmentByTag(getFragmentTag()) == null) {
                 CTInboxListViewFragment listView = new CTInboxListViewFragment();
                 listView.setArguments(bundle);
                 getSupportFragmentManager().beginTransaction()
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
@@ -1,6 +1,5 @@
 package com.clevertap.android.sdk.inbox;
 
-import android.content.Context;
 import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.graphics.drawable.Drawable;
@@ -244,13 +243,13 @@ public class CTInboxActivity extends FragmentActivity implements CTInboxListView
     }
 
     @Override
-    public void messageDidClick(Context baseContext, int contentPageIndex, CTInboxMessage inboxMessage, Bundle data,
+    public void messageDidClick(int contentPageIndex, CTInboxMessage inboxMessage, Bundle data,
                                 HashMap<String, String> keyValue, int buttonIndex) {
         didClick(data, contentPageIndex, inboxMessage, keyValue, buttonIndex);
     }
 
     @Override
-    public void messageDidShow(Context baseContext, CTInboxMessage inboxMessage, Bundle data) {
+    public void messageDidShow(CTInboxMessage inboxMessage, Bundle data) {
         Logger.v("CTInboxActivity:messageDidShow() called with: data = [" + data + "], inboxMessage = [" + inboxMessage .getMessageId()+ "]");
         didShow(data, inboxMessage);
     }
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
@@ -243,11 +243,14 @@ public class CTInboxActivity extends FragmentActivity implements CTInboxListView
      * updates, so an on-screen list never changes without a user gesture.
      */
     void refreshAllInboxListFragments() {
+        int refreshed = 0;
         for (Fragment fragment : getSupportFragmentManager().getFragments()) {
             if (fragment instanceof CTInboxListViewFragment && fragment.isAdded()) {
                 ((CTInboxListViewFragment) fragment).refreshList();
+                refreshed++;
             }
         }
+        Logger.v("refreshAllInboxListFragments: refreshed " + refreshed + " inbox fragment(s)");
     }
 
     @Override
--- a/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
+++ b/clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
@@ -8,7 +8,6 @@ import android.os.Bundle;
 import android.view.View;
 import android.widget.FrameLayout;
 import android.widget.LinearLayout;
-import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.RestrictTo;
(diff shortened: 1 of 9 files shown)


shubhambjadhavar:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clevertap-platform/pr-checks/9b7525a1-bdcc-4c5e-8f68-476fd31962ce?refs=eabe8a70becf19a520e6d3171f271abf091626cc&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clevertap-platform/pr-checks/9b7525a1-bdcc-4c5e-8f68-476fd31962ce/license?refs=eabe8a70becf19a520e6d3171f271abf091626cc&source=prChecksComment) |
| :white_check_mark: | Code Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/clevertap-platform/project/9a1fd2dd-3229-4564-b8c3-9660985efc7a/pr-check/ab7633bf-b9ae-4dee-8602-48e0e0e31457?refs=eabe8a70becf19a520e6d3171f271abf091626cc&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
```diff
@@ -122,37 +118,31 @@ public void onClick(View v) {
...
+                    fragmentExists = true;
                 }
             }
+            if (!fragmentExists) {
+                CTInboxListViewFragment listView = new CTInboxListViewFragment();
+                listView.setArguments(bundle);
+                getSupportFragmentManager().beginTransaction()
+                        .add(R.id.list_view_fragment, listView, getFragmentTag())
+                        .commit();
+            }
```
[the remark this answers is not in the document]
deeksha-rgb: @piyush-kukadiya

deeksha-rgb (approved):

clevertap-core/src/test/java/com/clevertap/android/sdk/inbox/CTInboxBaseMessageViewHolderTest.kt
```diff
@@ -0,0 +1,120 @@
...
+        val messageB = unreadMessage("msg_B")
+        holder.configureWithMessage(messageA, fragmentSpy, 5)
+        holder.invokeMarkItemAsRead(messageA, 5)
+
+        holder.configureWithMessage(messageB, fragmentSpy, 0)
+        fragmentSpy.inboxMessages.clear()
+
+        shadowOf(Looper.getMainLooper()).idleFor(Duration.ofMillis(2000))
+
+        verify(exactly = 0) { fragmentSpy.didShow(any(), any()) }
```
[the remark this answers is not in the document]
piyush-kukadiya: @coderabbitai skip this

clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxListViewFragment.java
```diff
@@ -224,6 +243,145 @@ public void onDestroy() {
...
+        if (currentMessages.size() != freshMessages.size()) {
+            return false;
+        }
+        for (int i = 0; i < currentMessages.size(); i++) {
+            CTInboxMessage current = currentMessages.get(i);
+            CTInboxMessage fresh = freshMessages.get(i);
+            if (!current.getMessageId().equals(fresh.getMessageId())
+                    || current.isRead() != fresh.isRead()
+                    || current.getDate() != fresh.getDate()
+                    || !contentJson(current).equals(contentJson(fresh))) {
```
CTLalit, line 364: we are already doing a string comparison, why do we need the preceding checks?

clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxListViewFragment.java
```diff
@@ -272,12 +455,12 @@ void didClick(Bundle data, int position, int contentPageIndex, HashMap<String, S
...
     @SuppressWarnings("SameParameterValue")
-    void didShow(Bundle data, int position) {
+    void didShow(Bundle data, CTInboxMessage inboxMessage) {
         CTInboxListViewFragment.InboxListener listener = getListener();
         if (listener != null) {
-            Logger.v("CTInboxListViewFragment:didShow() called with: data = [" + data + "], position = [" + position + "]");
+            Logger.v("CTInboxListViewFragment:didShow() called with: data = [" + data + "], messageId = [" + inboxMessage.getMessageId() + "]");
             //noinspection ConstantConditions
-            listener.messageDidShow(getActivity().getBaseContext(), inboxMessages.get(position), data);
+            listener.messageDidShow(getActivity().getBaseContext(), inboxMessage, data);
```
CTLalit: why use `getBaseContext` here?

clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
```diff
@@ -239,6 +226,23 @@ protected void onDestroy() {
     }
 
 
+    /**
+     * Repaints every resident inbox list fragment from the already-committed message
+     * store, so tabs never show contradicting data after a refresh. Called only from
+     * the pull-to-refresh success path (explicit user action) — never from background
+     * updates, so an on-screen list never changes without a user gesture.
+     */
+    void refreshAllInboxListFragments() {
```
CTLalit, line 234: we should just update the main dataset and only refresh the tabs once active? Also no other way to do this via adapter rather then a generic getFragments()

CTLalit (commented):
Posted some comments.

clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxListViewFragment.java
piyush-kukadiya: The string comparison only covers the `msg` part of the payload — the creative content (title, text, media, colors). It does not cover the whole message. A message actually looks like this:

```json
{
  "id": "1674122920_1674457409",   // outside msg
  "date": 1674457409,              // outside msg
  "isRead": 0,                     // outside msg
  "msg": { ...title, text, media... }   // <- only THIS is string-compared
}
```

So the three checks before it each catch something the string cannot see:

**1. `messageId` + `date`:** imagine marketing deletes a campaign and publishes it again with the exact same creative. The new message has the same `msg` content but a different id and date. Without these checks we would say "nothing changed" and skip the repaint — the user would keep seeing the old "2 hours ago" timestamp instead of "Just Now".

**2. `isRead` — this one cannot come from JSON at all:** when a user views a message for 2 seconds, the view holder flips the `isRead` *field* on the on-screen object — but that object's backing JSON still says `isRead: 0` (it was created at fetch time). The store write also happens async, so the fresh copy's JSON may say `0` or `1` depending on timing. Example: user reads a message, then pulls to refresh one second later. If we compared read state through JSON, the two copies would look different even though nothing really changed → we would repaint for no reason → a playing video gets paused. Comparing through the getter (after `mergeReadStateForward`) is the only version of read state that is always correct.

**3. Small bonus:** the three cheap checks run first, so on any row that already differs we never pay for the `JSONObject.toString()` serialization.

Comparing the full `data` JSON in one go fails for the same reason as 2 — the `isRead` key inside it is stale/volatile, so full-JSON equality gives wrong answers in both directions.

clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxListViewFragment.java
piyush-kukadiya: Good catch — this was pre-existing (passed here since the original inbox implementation, SDK-609); the line only appears in the diff because the parameter next to it changed for the stale-position fix.

Digging into it: the context argument was actually never consumed — `CTInboxActivity` is the only `InboxListener` implementation and it ignores the `baseContext` parameter in both callbacks. So instead of answering "why getBaseContext", removed the `Context` parameter from the (package-private) interface entirely.

Fixed in 3866c88 — interface + both call sites + `CTInboxActivity` overrides cleaned up. Bonus: this also removes the `getActivity().getBaseContext()` call whose potential null-deref was being suppressed with `//noinspection ConstantConditions`.

clevertap-core/src/main/java/com/clevertap/android/sdk/inbox/CTInboxActivity.java
CTLalit: this is by design, ignoring for now.

CTLalit (approved):

